PR #116, opened 2014-08-02
WIP: GtkApplication Experiments

tknopp:
Ok I had a look at `GtkApplication` and got a simple example running including a global menu on OSX. Yay.

But this raises questions how to move forward. Much of the functionality around `GtkApplication` is implemented in `GIO`. So is it possible to autogenerate the accessors for `GIO` types as well?

To get the menu actually working one has to implement the `GAction` interfaces. All this looks a little complicated and I wonder if one could define this using the regular `GtkMenu` in code (without `GtkBuilder`) with regular callbacks.

### DIFF
--- a/src/application.jl
+++ b/src/application.jl
@@ -10,6 +10,12 @@ if gtk_version == 3
 push!(action_map::GActionMap, action::GAction) = 
   ccall((:g_action_map_add_action, libgio), Void, (Ptr{GObject}, Ptr{GObject}), action_map, action)
 
+splice!(action_map::GActionMap, action_name::String) = 
+  ccall((:g_action_map_remove_action, libgio), Void, (Ptr{GObject}, Ptr{Uint8}), action_map, bytestring(action_name))
+
+lookup(action_map::GActionMap, action_name::String) = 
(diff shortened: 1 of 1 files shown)


malbert137:
First, thank you to all who looked into this.

I suspect it's easier generally to not use the "application" thing, but for what it's worth, this seems to work for me:

```
#!/usr/bin/env julia

import Gtk

function on_clicked(w)
  @info "clicked!" # println() doesn't work here?
end


function activate(app)
  window = Gtk.GtkApplicationWindow(app)
  Gtk.set_gtk_property!(window, :title, "Hello")

  button_box = Gtk.GtkBox(:v)
  push!(window, button_box)

  button = Gtk.GtkButton("Hello world!")
  push!(button_box, button)
  Gtk.signal_connect(on_clicked, button, :clicked)

  #Gtk.showall(app) #hangs here?
  #@info "showall() ok..?"

  Gtk.showall(window) # without this, no visible window
end

# execution starts here
app = Gtk.GtkApplication("stuff.fun.my", 
        Gtk.GConstants.GApplicationFlags.FLAGS_NONE)

Gtk.signal_connect(activate, app, :activate)

Gtk.run(app)

```

PR #126, opened 2014-10-01
add methods for Trees

jverzani:
Suggested additions to lists.jl for trees

# traversal

The main methods for traversal use the `GtkTreePath` interface:

copy(path::GtkTreePath) 
next(path::GtkTreePath) 
prev(path::GtkTreePath) 
up(path::GtkTreePath) 
down(path::GtkTreePath) 
string(path::GtkTreePath) 

These return a path (when working).

It is also convenient to work with iters. This pull request implements
the traversal methods for iters from
https://developer.gnome.org/gtk3/stable/GtkTreeModel.html

gtk_tree_model_get_iter_first ()
gtk_tree_model_iter_next ()           `next(model, iter)`
gtk_tree_model_iter_previous ()       `prev(model, iter)`
gtk_tree_model_iter_children ()       
gtk_tree_model_iter_has_child ()
gtk_tree_model_iter_n_children ()     `length(model, iter)`
gtk_tree_model_iter_nth_child ()      
gtk_tree_model_iter_parent ()         `parent(model, iter)`
gtk_tree_model_get_string_from_iter ()`string(model, iter)`

These generally get passed an iter that is mutated and return a
boolean indicating success. I don't know if that is the desired
interface for julia. It seem easier to not expose mutable iters and
instead pass an iter and return a tuple of (Bool, GtkTreeIter). 

On top of this, I added an iterator TreeIterator that implements a
depth first search over the children of an iter. This makes
traversal very easy. I gave the constructor the name `walktree`. Don't
really like it, but haven't thought of a better one. It is used like

```
for iter in Gtk.walktree(store)	# didn't export, but should once named properly
    key = GtkTreeModel(store)[iter, 1]
    println(key)
end
```

Or, `walktree(store, iter)` to walk the children of a given node.

* added `index_from_iter(GtkTreeModel, iter)` method to get easy to
  read index. Parallels `iter_from_index` for list stores.

* Added getindex calls for GtkTreeStore akin to those for
  GtkListStore. These pass off to a call with `GtkTreeModel(store)`. It doesn't seem like this should be necessary,
  but without this cast, getindex(store::GtkTreeStore, ...) will not dispatch to the method for GtkTreeModel.

* add 
  + `setindex!(store::Union(GtkListStore, GtkTreeStore), value, iter::Gtk.TRI, column::Integer)` 
  + `setindex!(store::GtkListStore, value, index::Int, column::Integer)` 
  + `setindex!(store::GtkTreeStore, value, index::Vector{Int}, column::Integer)` 

Calling the method specialized for GtkTreeModel doesn't work as there is no gtk_tree_model_set_value.

```
julia> iter = Gtk.get_iter_first(GtkTreeModel(store))
GtkTreeIter(...)

julia> GtkTreeModel(store)[iter, 1]
"ONE"

julia> GtkTreeModel(store)[iter, 1] = "one"

(<unknown>:6064): GLib-GObject-WARNING **: gvalue.c:179: cannot initialize GValue with type 'gchararray', the value has already been initialized as 'gchararray'
"one"
```

### DIFF
--- a/src/lists.jl
+++ b/src/lists.jl
@@ -464,18 +464,18 @@ index_from_iter(treeModel::GtkTreeModel, iter::TRI) = map(int, split(get_string_
 type TreeIterator
     store::GtkTreeStore
     model::GtkTreeModel
-    iter::Union(Nothing, GtkTreeIter)
+    iter
     next
 end
 
 function walktree(store::GtkTreeStore, iter=nothing; method::Symbol=:depth_first)
     model = GtkTreeModel(store)
 
-    if isa(iter, Nothing)
+    if iter === Nothing
         return TreeIterator(store, model, nothing, nothing)
     end
 
-    !Gtk.iter_has_child(model, iter) && error("Iter must be a parent")
+    Gtk.iter_has_child(model, iter) !! error("Iter must be a parent")
     TreeIterator(store, model, iter, nothing)
 end
   
--- a/src/lists.jl
+++ b/src/lists.jl
@@ -106,7 +106,7 @@ function iter_from_string_index(store, index::String)
     if !isvalid(store, iter)
         error("invalid index: $index")
     end
-     iter
+     iter[]
 end
 
 ### GtkListStore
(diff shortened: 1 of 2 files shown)


timholy:
Can this be closed?

vtjnash:
looks like it.

jverzani:
Sorry, closed now

PR #127, opened 2014-10-02
Pull request/d8fe3df2

jverzani:
This is similar to #126 but assumes the user will use mutable iters. I see that is how they are to use GtkTextIters and is the same as the gtk_ methods, so closer to the official documentation. Seems to me this is  just a bit more work but has the advantage of being more familiar and would be consistent with these methods if /when they are generated programmatically.


jverzani:
Bump. Has this fallen from the radar again? I'd be happy to put in a new pull request to avoid the conflicts.

PR #152, opened 2015-05-26
Correct deprecation on deps/build.jl

kaslusimoes:
Corrected the following deprecation issue on `Julia 0.4.0-dev+4968` commit `b85ede9`:

```julia
WARNING: deprecated syntax "[a=>b, ...]" at ...
Use "Dict(a=>b, ...)" instead.
```


prcastro:
The reason Travis failed seems unrelated

vtjnash:
committed as f56de95

tkelman:
This is 0.4-only syntax. Needs to use Compat to work on 0.3.

PR #160, opened 2015-08-23
Automatically precompile Gtk.jl for newer versions of Julia.

ufechner7:
Tested on 64 bit Linux. Reduced load time from 5.8 s to 1.1 s on Core i7, 3.4 GHz.


vtjnash:
Thanks. This package is the original and primary test case for `Base.cachecompile`, so it's certainly supposed to be working.

PR #162, opened 2015-08-28
Fix return value of signal_emit for void returns

timholy:


lobingera:
Just as i see this, where is this needed?

timholy:
The test portion of this fails on current master.

lobingera:
I was unclear: Is this used somewhere else, except testing? Afaiu g_signal_emit is used in inter-widget communication. Do we have a use-case for that?

timholy:
The main point is: it's a bugfix. You can't index an array with `Void`.

More detail on where this is useful: if you set the `:text` property of an `Entry` box, it doesn't fire the callback that would be fired if the user pressed `Enter` in the box. So this can be helpful if you're manipulating the entry box programmatically. See https://github.com/timholy/GtkUtilities.jl#managing-state

PR #163, opened 2015-08-29
Improve documentation on signals

timholy:
Motivated by https://github.com/JuliaLang/Gtk.jl/issues/161#issuecomment-135972461


vtjnash:
thanks

PR #169, opened 2015-09-09
Implement stack operations for mouse handler callbacks, and add scroll

timholy:
From the commit message:
Gtk calls signal handlers in the order they are defined. Since
the handlers associated with MouseHandler are defined first,
they are always the first to run. However, sometimes you want a
later-defined callback to temporarily replace an existing callback.
To achieve this functionality, we replace the current callback and
keep the old one on a stack.

A single stack is used for all event types (e.g., button1press,
button2release, etc). Given that this stack
probably won't grow large, it would seem vaguely silly to introduce
9 new fields for storing the stack for each separate event.

This also adds a scroll callback, to provide MouseHandler support for wheelmice.


timholy:
The only test failure here is on OSX where we lack the equivalent of `xvfb-run`.

PR #174, opened 2015-09-14
Add input_dialog

timholy:
```jl
resp, firstname = input_dialog("Enter your first name, if you want to", "Julia", (("No thanks",0), ("OK",1)))
```
image


jiahao:
Why is the default icon not three colored circles? ;-)

timholy:
That is a big flaw indeed. The topic of the next JuliaCon hackathon?

PR #176, opened 2015-09-15
add basic GtkClipboard support

jonathanBieler:
It's the first time I do this, I hope I didn't messed up somewhere. I didn't added short names because there's already some clipboard methods in Base, but maybe it's not a problem.

`clipboard_wait_for_text` crashes when it's called from the terminal because Gtk checks for a `main_loops` that's not defined in that context, is there a way to check that from Julia to make it safer?

I tested it only on windows 7.


src/text.jl
```diff
@@ -83,10 +83,22 @@ immutable GtkTextRange <: Range
...
-#TODO
-#end
-
+#####  GtkClipboard #####
+
+GtkClipboardLeaf(selection::UInt16) =  GtkClipboardLeaf(ccall((:gtk_clipboard_get,libgtk), Ptr{GObject},
+    (UInt16,), selection))
+GtkClipboardLeaf() = GtkClipboardLeaf(Gtk.GdkAtoms.CLIPBOARD)
+clipboard_set_text(clip::GtkClipboard,text::String) = ccall((:gtk_clipboard_set_text,libgtk), Void,
+    (Ptr{GObject}, Ptr{Uint8},Cint), clip, text, length(text))
```
vtjnash: `sizeof(text)`, not length

vtjnash: I believe this will be exposed automatically as `G_.text(clipboard, text)` once the autogen script is rerun.

src/gdk.jl
```diff
@@ -68,6 +68,86 @@ baremodule GdkKeySyms
   const Hyper_R = 0xffee
 end
 
+baremodule GdkAtoms
```
vtjnash: where are you getting all these from? Gtk itself appears to only defines "GDK_NONE"

PyGTK defines a few extras (https://developer.gnome.org/pygtk/stable/class-gdkatom.html), but only because it has computed the value at runtime (https://developer.gnome.org/gdk3/stable/gdk3-Properties-and-Atoms.html#gdk-atom-intern)

jonathanBieler:
I couldn't find the value of the CLIPBOARD one anywhere so I generated the list by calling `gdk_atom_name`. Thinking about it it's probably not the best idea, these are maybe not constant. 

Some are defined here though: 

https://developer.gnome.org/gdk3/stable/gdk3-Selections.html#GDK-SELECTION-PRIMARY:CAPS

I can restrict them to this list. Otherwise I just saw there's `gdk_atom_intern` that allows to get the value at runtime, it's maybe safer.

src/long_leaf_exports.jl
```diff
@@ -134,6 +135,7 @@ export
     GtkTextMarkLeaf,
     GtkTextTagLeaf,
     GtkTextViewLeaf,
+    GtkClipboardLeaf,
     GtkToolButtonLeaf,
     GtkToolItemLeaf,
     GtkToolbarLeaf,
```
vtjnash: please make the corresponding updates to `short_leaf_exports.jl` and `short_exports.jl`

src/gdk.jl
```diff
@@ -68,6 +68,21 @@ baremodule GdkKeySyms
...
+    const SELECTION_TYPE_ATOM = 0x0004
+    const SELECTION_TYPE_BITMAP = 0x0005
+    const SELECTION_TYPE_COLORMAP = 0x0007
+    const SELECTION_TYPE_DRAWABLE = 0x0011
+    const SELECTION_TYPE_INTEGER = 0x0013
+    const SELECTION_TYPE_PIXMAP = 0x0014
+    const SELECTION_TYPE_STRING = 0x001f
+    const SELECTION_TYPE_WINDOW = 0x0021
+    const SELECTION_CLIPBOARD = 0x0045
+end
```
vtjnash: The type `GdkAtom` is a `Ptr{Void}`. It'll be clearest to define `typealias GdkAtom Ptr{Void}`, and use that to create these and for below.

LilithHafner:
Does this have any advantages over `InteractiveUtils.clipboard`?

PR #179, opened 2015-09-17
Fix binding deprecations and bump julia version

timholy:
With the new binding deprecation warnings, this package has gotten very noisy. This fixes the problem on julia 0.4. julia 0.3 is more problematic, particularly because we can't easily use the `Compat` package due to an internal module called `Compat`. So I'm proposing the simple expedient of switching master to require julia 0.4-. If you like that idea, I can separately push a julia-0.3 branch.

### DIFF
--- a/src/GLib/gtype.jl
+++ b/src/GLib/gtype.jl
@@ -23,14 +23,14 @@ const fundamental_types = (
     (:gchar,      Int8,             Int8,           :schar),
     (:guchar,     UInt8,            UInt8,          :uchar),
     (:gboolean,   Cint,             Bool,           :boolean),
-    (:gint,       Cint,             Union(),           :int),
-    (:guint,      Cuint,            Union(),           :uint),
-    (:glong,      Clong,            Union(),           :long),
-    (:gulong,     Culong,           Union(),           :ulong),
+    (:gint,       Cint,             Union{},           :int),
+    (:guint,      Cuint,            Union{},           :uint),
+    (:glong,      Clong,            Union{},           :long),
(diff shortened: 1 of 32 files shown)


src/GLib/gtype.jl
```diff
@@ -18,22 +18,22 @@ end
...
+    (:guchar,     UInt8,            UInt8,          :uchar),
     (:gboolean,   Cint,             Bool,           :boolean),
-    (:gint,       Cint,             None,           :int),
-    (:guint,      Cuint,            None,           :uint),
-    (:glong,      Clong,            None,           :long),
-    (:gulong,     Culong,           None,           :ulong),
+    (:gint,       Cint,             Union(),           :int),
+    (:guint,      Cuint,            Union(),           :uint),
+    (:glong,      Clong,            Union(),           :long),
+    (:gulong,     Culong,           Union(),           :ulong),
```
vtjnash: `Union{}`

yuyichao: `Union()` is not giving deprecation warning yet. Should it?

vtjnash:
Branching sounds good to me. I want to drop the `@` and make use of call-overloading shortly after 0.4 is released anyways.

src/GLib/gtype.jl
timholy: I used `Union()` because initially I was striving for julia 0.3 compatibility. But I can change that.

> Union() is not giving deprecation warning yet. Should it?

Probably worth filing a julia issue about it.

timholy:
OK, fixed.

Travis seems backlogged ATM, so it may be a wait.

PR #182, opened 2015-09-23
Set additional gdk-pixbuf env vars for Homebrew

staticfloat:
This combined with https://github.com/staticfloat/homebrew-juliadeps/pull/73 should fix https://github.com/JuliaGraphics/Immerse.jl/issues/11.


vtjnash:
thanks!

PR #183, opened 2015-09-24
Union(x) -> Union{x}

timholy:


timholy:
There are still a few depwarns from stuff that's in `gen`. I'm a little vague on what you do to regenerate that.

PR #191, opened 2015-10-12
Typo in example of warn_dialog

amellnik:


vtjnash:
this wasn't necessarily a typo, but an example of how you can pass a second parameter to be the owner window. it would be nice to know if that is missing its mark.

amellnik:
Ah whoops, I missed the "optional parent container" bit while trying out the code snippets from the readme.  Perhaps this could be made clear by splitting this out into a code block than can be run as is.  A cheesy example might be:

#### Message dialogs

Gtk.jl also supports `GtkMessageDialog` and provides several convenience functions:  `info_dialog`, `ask_dialog`, `warn_dialog`, and `error_dialog`.  Each inputs a string and returns nothing, except for `ask_dialog` which returns true if the user clicked `yes`.


```jl
info_dialog("Julia rocks!")
ask_dialog("Do you like chocolate ice cream?", "I like it", "Not at all") && println("That's my favorite too.")
warn_dialog("Oops!... I did it again")
```

It is also possible to make an existing window the parent by passing it as an optional second parameter.

```jl
win = @Window("Interacting with bears") |> (f = @Frame("Would you like to..."))
vbox = @Box(:v)
push!(f, vbox)
feed = @Button("Feed them")
safe = @Button("Keep a safe distance")
push!(vbox, feed)
push!(vbox, safe)
showall(win)
warn_dialog("Please don't feed the wildlife!", win)
```

PR #197, opened 2015-11-10
Disable show method for types until julia#13306 is resolved

timholy:


timholy:
Failure is a travis script problem fixed in #196

PR #204, opened 2015-11-29
Fix CssProviderLeaf(filename="...")

fcard:
There was an extra type in the c function signature.

### DIFF
--- a/src/theme.jl
+++ b/src/theme.jl
@@ -15,7 +15,7 @@ function GtkCssProviderLeaf(;data=nothing,filename=nothing)
     elseif filename !== nothing
         GError() do error_check
           ccall((:gtk_css_provider_load_from_path,libgtk), Bool,
-            (Ptr{GObject}, Ptr{UInt8}, Clong, Ptr{Ptr{GError}}),
+            (Ptr{GObject}, Ptr{UInt8}, Ptr{Ptr{GError}}),
             provider, bytestring(filename), error_check)
         end
     end
--- a/test/gui.jl
+++ b/test/gui.jl
@@ -564,3 +564,25 @@ str = takebuf_string(io)
 @assert str == "scrolling\n"
 
 destroy(w)
+
+# CSS
+
+## CssProviderLeaf(filename="...")
+
+style_file = "style_test.css"
+
+l = @Label "I am some large blue text!"
+w = @Window(l)
+
+screen   = Gtk.GAccessor.screen(w)
+provider = CssProviderLeaf(filename=style_file)
+
+ccall((:gtk_style_context_add_provider_for_screen, Gtk.libgtk), Void,
+      (Ptr{Void}, Ptr{GObject}, Cuint),
+      screen, provider, 1)
+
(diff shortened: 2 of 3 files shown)


vtjnash:
the return type is wrong too -- should be `Cint != 0`, not `Bool` (also for gtk_css_provider_load_from_data)

https://developer.gnome.org/gtk3/stable/GtkCssProvider.html#gtk-css-provider-load-from-path

vtjnash:
also, can I ask you to squash the commits into one commit?

fcard:
I will, but for some reason the travis log keeps saying the css file is invalid (even through it works for me both on windows and linux), so I am trying to guess what the problem is and fix that.

fcard:
Yeah, I don't get it, for some reason travis is reading the css file differently. I am going to try to reproduce the travis environment here to debug this out, I will be back soon.

Meanwhile, `CssProviderLeaf(data="...")` locks my computer on windows, no clue about that one either.

fcard:
Done! There was some problem with travis's gtk when parsing the `font-size` attribute.
@vtjnash, changing the return type caused an method error with `!`, so I am going to leave it as is for this PR, if you don't mind, to keep things simple.

hustf:
A checkout (with this included) fixed:

```
(julia.exe:447752): GModule-CRITICAL **: g_module_symbol: assertion 'symbol != N
ULL' failed
```
which, I believe, appears when calling win = Gtk.@Window("stuff", 800,600).

I believe this lead to unexpected behaviours in recent additions GTK utilities, Immerse and ProfileView as a window is actually opening and appears to work fine. This error message from GTK is already scrolled way up the screen when errors occur. So tag, please.

hustf:
Sorry, I was too quick there. The specific test case worked. I jumped into other modules and that may have changed some state. If I run Pkg.test("Gtk") on version 0.9.3+ I get:

´´´
julia> Pkg.test("Gtk")
INFO: Testing Gtk
.juliarc define function: lastfile()

(julia.exe:449068): GModule-CRITICAL **: g_module_symbol: assertion 'symbol != NULL' failed
**(repeated around 30 times)**
WARNING: Method definition g_type(Type{##9#T<:Gtk.GtkWidget}) in module Gtk at C:\Users\F\.julia\v0.4\Gtk\src\GLib\
gtype.jl:181 overwritten in module Test at C:\Users\F\.julia\v0.4\Gtk\src\GLib\gtype.jl:181.
The following messages:
   WARNING: Error in @guarded callback
are expected and a sign of normal operation.
WARNING: Error in @guarded callback
ERROR: UndefVarError: k not defined
WARNING: Error in @guarded callback
ERROR: UndefVarError: k not defined
WARNING: Error in @guarded callback
ERROR: UndefVarError: k not defined
WARNING: Error in @guarded callback
ERROR: UndefVarError: k not defined

(julia.exe:449068): GModule-CRITICAL **: g_module_symbol: assertion 'symbol != NULL' failed
INFO: Gtk tests passed  (edit: :blush: )

PR #212, opened 2015-12-27
simplify sigatom code

vtjnash:
this doesn't fix 161, it just moves the failure condition so that it
should be much less common and more avoidable

### DIFF
--- a/src/GLib/signals.jl
+++ b/src/GLib/signals.jl
@@ -66,35 +66,35 @@ function GClosureMarshal(closuref, return_value, n_param_values,
     cb = unsafe_load(closure_env, 1)
     gtk_calling_convention = (0 != unsafe_load(convert(Ptr{Int},closure_env), 2))
     params = Array(Any, n_param_values)
-    if gtk_calling_convention
-        # compatibility mode, if we must
-        param_types,user_data = unsafe_load(closure_env, 2)::Array{Any,1}
-        length(param_types)+1 == n_param_values || error("GCallback called with the wrong number of parameters")
-        for i = 1:n_param_values
-            gv = mutable(param_values,i)
-            gtyp = unsafe_load(gv).g_type
-            # avoid auto-unboxing for some builtin types in gtk_calling_convention mode
-            if g_isa(gtyp,g_type(GObject))
-                params[i] = ccall((:g_value_get_object,libgobject), Ptr{GObject}, (Ptr{GValue},), gv)
-            elseif g_isa(gtyp,g_type(GBoxed))
-                params[i] = ccall((:g_value_get_boxed,libgobject), Ptr{Void}, (Ptr{GValue},), gv)
-            elseif g_isa(gtyp,g_type(AbstractString))
-                params[i] = ccall((:g_value_get_string,libgobject), Ptr{Void}, (Ptr{GValue},), gv)
-            else
-                params[i] = gv[Any]
-            end
-            if i > 1
-                params[i] = convert(param_types[i-1], params[i])
-            end
-        end
-        push!(params, user_data)
-    else
-        for i = 1:n_param_values
-            params[i] = mutable(param_values,i)[Any]
-        end
-    end
     local retval = nothing
     g_siginterruptible(cb) do
+        if gtk_calling_convention
+            # compatibility mode, if we must
+            param_types,user_data = unsafe_load(closure_env, 2)::Array{Any,1}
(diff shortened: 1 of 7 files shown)


src/GLib/signals.jl
```diff
@@ -146,39 +146,39 @@ function signal_emit(w::GObject, sig::AbstractStringLike, RT::Type, args...)
...
 gtk_stack = nothing # need to call g_loop_run from only one stack
-gtk_work = nothing
-g_sigatom_flag = false
-function g_sigatom(f::Base.Callable)
-    # don't yield() from f inside g_sigatom, use g_siginterruptible(yield) if unsure
-    global g_sigatom_flag, gtk_stack, gtk_work
+gtk_yielded = false # when true, use the `gtk_doatomic` queue to run sigatom functions
+gtk_doatomic = [] # (work, notification) scheduler queue
+g_sigatom_flag = false # keep track of Base sigatomic state
+function g_sigatom(f::Base.Callable) # calls f, were f never throws (but this function may throw)
```
timholy: were->where

src/GLib/signals.jl
```diff
@@ -146,39 +146,39 @@ function signal_emit(w::GObject, sig::AbstractStringLike, RT::Type, args...)
...
+        push!(gtk_doatomic, (f, ct))
+        return wait()
+    end
+
     if !prev
         sigatomic_begin()
         g_sigatom_flag = true
     end
-    local ret
+    local ret = nothing
```
timholy: don't need the `local` now

vtjnash:
This should be better now and make #161 happier

vtjnash:
So, even though this PR doesn't fix #161, it makes it more trivial to work around:

1. if you are in a callback, don't yield. use `@async print`, etc, or queue a message for yourself, etc. You can also call `GLib.g_yield()`, but see 2

2. if you are getting segfaults in some random method due to there existing a callback that recursively calls the glib main loop (such as making a dialog box) or otherwise causes `g_yield` to be called, wrap the faulting code in `@sigatom`. This will postpone execution of the code until it can be run on the proper stack (but otherwise acts like normal control flow).

src/GLib/signals.jl
```diff
@@ -151,165 +151,113 @@ function signal_emit(w::GObject, sig::AbstractStringLike, RT::Type, args...)
...
-gtk_work = nothing
-g_sigatom_flag = false
-function g_sigatom(f::Base.Callable)
-    # don't yield() from f inside g_sigatom, use g_siginterruptible(yield) if unsure
-    global g_sigatom_flag, gtk_stack, gtk_work
-    prev = g_sigatom_flag
-    stk = gtk_stack
+g_stack = nothing # need to call g_loop_run from only one stack
+const g_yielded = Ref(false) # when true, use the `g_doatomic` queue to run sigatom functions
+const g_doatomic = [] # (work, notification) scheduler queue
```
timholy: Is there any virtue in declaring this `Tuple{Any,Any}[]`?

timholy:
Maybe add a version of those two comments to the README?

:100:

src/GLib/signals.jl
vtjnash: I suspect that would be slower

timholy: Fine then.

timholy:
Awesome. Tag a new release?

vtjnash:
I figured I would give it a day or two for the bug reports to start coming in. But if it seems to be working, then sure, I can tag a new release.

timholy:
I haven't tested it yet myself, but that may change by the end of the day.

Thanks for tackling this, I'm really looking forward to it.

timholy:
I tested whether I could live without https://github.com/timholy/GtkUtilities.jl/pull/7 by wrapping
https://github.com/timholy/GtkUtilities.jl/blob/9ef4557c7531975ccde2f7e245596588bcfc939c/src/guidata.jl#L42-L48 in `@sigatom` and/or `@schedule`. It's much better, but I still can get the segfault.

If you need some test code, I can arrange something. That said, I know you have many priorities, so no rush.

vtjnash:
Test code would be good. There shouldn't be any difference between guarded and siginterruptable now

tknopp:
@vtjnash: I have tested it during last week and `sigatom` made all my segfaulting apps just work. This is very much appreciated. Would be great if a new release could be tagged.

PR #214, opened 2016-01-19
added gtk_grid_remove_row/column convenience function

bjarthur:
this complements the already existing `insert!` for grids with a corresponding `deleteat!`

### DIFF
--- a/src/layout.jl
+++ b/src/layout.jl
@@ -57,6 +57,16 @@ if gtk_version == 3
         end
     end
 
+    function deleteat!(grid::GtkGrid, i::Integer, side::Symbol)
+        if side == :row
+            ccall((:gtk_grid_remove_row,libgtk), Void, (Ptr{GObject}, Cint), grid, i)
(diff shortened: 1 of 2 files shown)


bjarthur:
now with tests

bjarthur:
checks possibly fail because gtk 3.10 is needed.  see https://github.com/JuliaLang/Gtk.jl/pull/209 for similar problem.

https://developer.gnome.org/gtk3/stable/api-index-3-10.html

tknopp:
@bjarthur Could you please add a version check as introduced in https://github.com/JuliaGraphics/Gtk.jl/pull/209/commits/7eaea7b1f5b460c15ae00fe347a5f0bd862143c9#diff-3adbb7231f221897af7c35696d3d98caR170
so that CI is happy?

bjarthur:
done!  thanks for asking!!

tknopp:
LGTM

tknopp:
LGTM

Issue #218, opened 2016-03-05
path_at_pos problem


lucianolorenti:
I'm having trouble with the function path_at_pos
`function path_at_pos(treeView::GtkTreeView, x::Integer, y::Integer)`

for some reason this lines fails
`path = convert(Gtk.GtkTreePath,pathPtr[])`

pathPtr is defined as 
`pathPtr = Gtk.mutable(Ptr{Gtk.GtkTreePath})`

The error message is:

> ERROR: MethodError: `convert` has no method matching convert(::Type{Ptr{Gtk.GtkTreePath}}, ::Void)
> This may have arisen from a call to the constructor Ptr{Gtk.GtkTreePath}(...),
> since type constructors fall back to convert methods.
> Closest candidates are:
>   call{T}(::Type{T}, ::Any)
>   convert{T}(::Type{Ptr{T}}, ::UInt64)
>   convert{T}(::Type{Ptr{T}}, ::Int64)
>   ...
> ERROR (unhandled task failure): MethodError: `convert` has no method matching convert(::Type{Ptr{Gtk.GtkTreePath}}, ::Void)
> This may have arisen from a call to the constructor Ptr{Gtk.GtkTreePath}(...),
> since type constructors fall back to convert methods.
> Closest candidates are:
>   call{T}(::Type{T}, ::Any)
>   convert{T}(::Type{Ptr{T}}, ::UInt64)
>   convert{T}(::Type{Ptr{T}}, ::Int64)
>   ...
> 
>

lucianolorenti:
This an example of what i am talking about. I copied the path_at_pos function into this code. I am not sure why isn't working. I've tried some workarounds but I couldn't make it works.
```
using Gtk

function _path_at_pos(treeView::GtkTreeView, x::Integer, y::Integer)
    pathPtr = Gtk.mutable(Ptr{Gtk.GtkTreePath})
    path = Gtk.GtkTreePath()

    ret = ccall((:gtk_tree_view_get_path_at_pos,Gtk.libgtk),Cint,
                      (Ptr{GObject},Cint,Cint,Ptr{Ptr{Void}},Ptr{Ptr{Void}},Ptr{Cint},Ptr{Cint} ),
                       treeView,x,y,pathPtr,C_NULL,C_NULL,C_NULL) != 0
    if ret
      path = convert(Gtk.GtkTreePath, pathPtr[])
    end
    ret, path
end 

[...]

PR #220, opened 2016-03-09
Fixed path_at_pos problem

lucianolorenti:
#218


src/lists.jl
```diff
@@ -84,7 +84,7 @@ type GtkTreePath <: GBoxed
         finalizer(path, x::GtkTreePath->begin
                 ccall((:gtk_tree_path_free,libgtk),Void,(Ptr{GtkTreePath},),x.handle)
             end)
-        path
+        x
```
vtjnash: it looks like the finalizer also used `path`?

lucianolorenti: you are right

PR #225, opened 2016-03-25
fixed gtk_tree_path_copy return value type

lucianolorenti:
In the GtkTreePath constructor the return type of gtk_tree_path_copy was Void but the function returns a pointer to GtkTreePath


vtjnash:
It seems you've pretty much taken over the active development of this project, so I've given you access to this repository so you can work directly here and have control of its future direction. Welcome!

PR #227, opened 2016-04-02
fix parsing on 0.5

### LINKED ISSUE #226: Gtk fails to parse on master.
The problem is that this syntax https://github.com/JuliaLang/Gtk.jl/blob/e6b8e49ed76b8fc92e7e3ff5781fec91b8102cc2/src/events.jl#L146 is now a parser error.

Is this statement even expressible in the 0.5 syntax rules?

One way to fix it is just to drop support for 0.3.

KristofferC:
cc @yuyichao 

Fixes #226

### DIFF
--- a/src/events.jl
+++ b/src/events.jl
@@ -143,7 +143,7 @@ type MouseHandler
 end
 
 if VERSION < v"0.4.0-dev"
-typealias MHPair  (MouseHandler,Symbol)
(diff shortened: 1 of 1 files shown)


KristofferC:
bump

Keno:
sure, LGTM

PR #231, opened 2016-05-10
The big julia-0.5 update

timholy:
Most of this is tedious, but there are a couple that might be worth a look:
- d9048f5506fd3b5b3c6f1fcd8614566db7c78591: I was lazy and didn't check whether there is a better way
- 11ae24d1228cc3cc254e4d385e60822a46786efc: is it weird this didn't cause an error before?

The one thing left to make this perfect :smile: on julia-0.5 would be to eliminate the overwritten function warnings in `gen/`.

### DIFF
--- a/src/GLib/GLib.jl
+++ b/src/GLib/GLib.jl
@@ -1,5 +1,9 @@
 module GLib
 
+if VERSION < v"0.5.0-dev"
+    include("../compat_string.jl")
+end
+
 if false
 function include(x)
     println("including $x")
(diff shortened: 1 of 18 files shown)


timholy:
Another point worth making: start a julia session and run the tests. Then exit. On julia-0.4, exiting happens "instantly." On julia-0.5, there's a ~3.5 second delay before you get back to the linux prompt. Calling `gc()` before exiting does not materially decrease this time.

timholy:
Given the travis backlog for JuliaLang, any objections if we move this package to https://github.com/JuliaGraphics?

src/GLib/GLib.jl
```diff
@@ -1,5 +1,9 @@
 module GLib
 
+if VERSION < v"0.5.0-dev"
```
tkelman: be more specific about version number cutoffs?

timholy: I'm just fitting in with the rest of the code :smile:.

tknopp:
Thanks for working on this! I am still on 0.4 but will appreciate the smooth transition due to your effort

timholy:
Of course, the `f.(args)` merger just introduced a whole bunch of new deprecations :smile:, but I'll wait until https://github.com/JuliaLang/Compat.jl/pull/196 merges and gets tagged.

PR #233, opened 2016-05-17
Add `empty!(combobox)`

timholy:
This empties the choices in a `ComboBoxText` so that you can repopulate it. I'd like to write a test, but I'm not quickly discovering how to iterate over the current items in the ComboBox. Any pointers?


tknopp:
maybe using `gtk_combo_box_get_model ()` and within that there seem to be a number of items function

timholy:
Thanks for the tip. Updated.

tknopp:
@timholy LGTM. I would merge but somehow lost write access due to the move to JuliaGraphics

PR #235, opened 2016-05-26
Make it easy to construct a dummy `GdkEventKey`, plus fix new deprecations

timholy:
I didn't fix the `@static` deprecation. There's a fix here:

https://github.com/JuliaLang/Compat.jl/commit/f59536084236dfa0437601111a20aab4c82cfba8

I'm increasingly wondering about the wisdom of avoiding Compat.jl.

### DIFF
--- a/deps/ext.jl
+++ b/deps/ext.jl
@@ -6,8 +6,14 @@ if isfile(_depspath)
     include(_depspath)
 end
 
+if VERSION >= v"0.5.0-dev+4257"
+    const KERNEL = Base.Sys.KERNEL
+else
+    const KERNEL = Base.OS_NAME
+end
+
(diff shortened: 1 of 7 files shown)


vtjnash:
> I'm increasingly wondering about the wisdom of avoiding Compat.jl.

probably not much anymore. initially, Compat.jl didn't exist / was very limited in version support. Gtk.jl then supported a wider range of Julia versions than Compat. Such is not the case today.

(`bytestring` is a different story though, since it is a similar, but not compatible, layer on top of the function Base of the same name)

timholy:
Agreed. I'll merge this and then think about others later. (So many deprecations, so little time :wink:...)

PR #236, opened 2016-05-26
WIP: Fix #134

### LINKED ISSUE #134: Use new call overloading
@vtjnash: Hi Jameson, if I get it right with Julia 0.4 we can have constructors even for abstract types. Do you think it would be feasible to get rid of the `@` names again? Or did I get this wrong...

tknopp:
This allows to create Gtk widgets without using macro syntax. The macro syntax is still there in order to have a smooth transition.
Thoughts @vtjnash @timholy ?


timholy:
Can I recommend you also get rid of the `@`s in the rest of `src` and `test`? Some will need to stay (`@assert`, `@guarded`, `@sigatom`, etc.), but presumably any `@` in front of a capital letter can be automatically dropped (`sed` would be your friend here, if you don't have another way).

timholy:
Also need exports for the non-`@` versions.

tknopp:
First problem. The type `Box` already exists in Julia Core...

tknopp:
Thanks Tim. So I ported the tests and despite the ShortName of GtkBox not being valid it seems to work.

So before I add all the exports for the `non-@` versions: should the `@`-exports stay there? Should we deprecate them?

I know that (once upon a time) where `@` was introduced it was done without deprecations. but simply by a rename. But this might not be ok today anymore (because of an hopefully increased user base).

vtjnash:
> First problem. The type Box already exists in Julia Core...

That shouldn't matter:
```julia
julia> Box
ERROR: UndefVarError: Box not defined
 in eval(::Module, ::Any) at ./boot.jl:225
 in (::Base.REPL.##1#2{Base.REPL.REPLBackend})() at event.jl:92
```

tknopp:
@vtjnash: On this branch:

```
julia> using Gtk.ShortNames
INFO: Precompiling module Gtk...
2016-05-27 14:57:40.777 julia[26348:4269110] *** WARNING: Method userSpaceScaleFactor in class NSView is deprecated on 10.7 and later. It should not be used in new applications. Use convertRectToBacking: instead. 

julia> Box
WARNING: both ShortNames and Core export "Box"; uses of it in module Main must be qualified
ERROR: UndefVarError: Box not defined

julia> 
```

Help?

timholy:
> So before I add all the exports for the non-@ versions: should the @-exports stay there? Should we deprecate them?

Yes, let's deprecate them.

tknopp:
I still have an issue with `GtkRadioButtonGroup` wich seems to be not exported

@timholy deprecation is fine. However, I am a newby in that area. Tried:
```
@deprecate @GtkWindow GtkWindow
```
but no luck.

timholy:
Try `@deprecate_binding`.

tknopp:
Ok, I learned about `depwarn` which made it possible to do this at a central place.

> Also need exports for the non-@ versions.

No these are already exported since the interfaces (abstract types) were used before. This PR just adds a `call` constructor to the abstract types that forwards to the leaf type.

So whats IMHO missing:

  - whats the issue with `GtkRadioButtonGroup`
  - what to do about ShortName `Box`
  - change all documentation

Unfortunately I don't speak `sed`...

vtjnash:
Maybe the short name only will work with master?

tknopp:
Which master? Julia master? I have not checked so far as I am on 0.4 currently.

If that is the case we could make the switch to the non-macro syntax once 0.5 is released.

tknopp:
this is now passing on Julia 0.5.

tknopp:
I have fixed it on Julia 0.4 and would merge this if there are no objections.
Would like to address #274 next.

PR #237, opened 2016-05-26
Generate libgtk_version during initialization

tknopp:
Implement the suggestion https://github.com/JuliaGraphics/Gtk.jl/pull/209#issuecomment-221935860

Will be necessary to conditionally implement things that have been introduced to Gtk during the version 3 cycle.


tknopp:
thanks

PR #240, opened 2016-06-17
fixed libgtk_version for gtk2

bjarthur:
this is at least a partial fix for gtk2.  specifically it fixes the determination of `libgtk_version`.  i can now `using Gtk` on a linux box with gtk v2.24.23 without any problems.

on a second box though, with v2.18.9, i also have to modify `deps/build.jl` to include `libgtk-x11-2.0` as a dependency.  not sure why this isn't necessary on the first box.  this version of gtk is perhaps not worth supporting as `GtkSpinner` is in `Gtk.jl` and only present in `gtk` since v2.20.

### DIFF
--- a/src/Gtk.jl
+++ b/src/Gtk.jl
@@ -40,10 +40,17 @@ include("long_exports.jl")
 include("long_leaf_exports.jl")
 include(joinpath("..","deps","ext.jl"))
 
-const libgtk_version = VersionNumber(
-    ccall((:gtk_get_major_version,libgtk),Cint, ()),
-    ccall((:gtk_get_minor_version,libgtk),Cint, ()),
-    ccall((:gtk_get_micro_version,libgtk),Cint, ()))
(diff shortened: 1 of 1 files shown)


tknopp:
maybe better check for the version instead of making a try catch.

bjarthur:
you would have me parse `libgtk` to figure out whether it's version 2 or 3?  that seems much more complicated than a simple try/catch.  or did you have something else in mind?

tknopp:
This is already used in the code, so my comment was on consistency:

https://github.com/JuliaGraphics/Gtk.jl/blob/a1cf0af804a5323f87e1a063b651d97562d40d68/src/theme.jl#L1

bjarthur:
doh!  forgot about that variable.  okay.  change made.  thanks!

tknopp:
Thanks for the fix!

PR #241, opened 2016-06-19
Add screen_geometry to get the screen size

timholy:


tknopp:
maybe calling it `screen_size`? 

Its not totally clear what the function that takes a window will do. Isn't the screen size always the same?

timholy:
Not if you have multiple displays.

tknopp:
I see, clever!

timholy:
libgtk seems to use `geometry` so I went with that. But I'm totally happy to change it to `screen_size` if folks prefer it.

tknopp:
not sure, I just stumbled upon this when reading https://github.com/JuliaGraphics/Gtk.jl/pull/241/files#diff-21a65f742951b3e1dc0ae1a3f0ca6418R12

So if this term is used in libgtk that is certainly fine.

timholy:
Hmm, seems better to change it to `screen_size`. Thanks!

Issue #194, opened 2015-11-06
Not building with gtk2


micahjsmith:
On a RHEL6.7 system, I only have gtk2 available. I tried modifying `deps/ext.jl` as instructed
```
const gtk_version = 2
```
But the build still fails:
```
julia> Pkg.build("Gtk")
INFO: Building Cairo
INFO: Building Gtk
===============================================================================[ ERROR: Gtk ]================================================================================

LoadError: None of the selected providers could satisfy library group gtk
Use BinDeps.debug(package_name) to see available providers

while loading /home/<myname>/.julia/v0.4/Gtk/deps/build.jl, in expression starting on line 58

=============================================================================================================================================================================

==============================================================================[ BUILD ERRORS ]===============================================================================

WARNING: Gtk had build errors.

 - packages with build errors remain installed in /home/<myname>/.julia/v0.4
 - build the package(s) and all dependencies with `Pkg.build("Gtk")`
 - build a single package by running its `deps/build.jl` script

=============================================================================================================================================================================
```

Does something get cached in between when I download the package and when I am able to modify `deps/ext.jl` after the failed build? Can't tell if I should be doing something differently or if this is a bug.

vtjnash:
the build script probably can't handle gtk2, but it shouldn't matter, since the dependency is already installed

micahjsmith:
Following up on this, if I run
```
julia> BinDeps.debug("Gtk")
```
I see
```
...
     - Library "gtk"
        - Providers:
          - BinDeps.AptGet package libgtk-3-0 (can't provide)
          - BinDeps.Yum package gtk3 (can't provide)
...
```

So it seems the build script is not picking up on the gtk_version variable.

tknopp:
I once had tried this and had issues in the actual code not working #140

bjarthur:
try adding `libgtk-x11-2.0` and `ibgdk-x11-2.0` to deps/builds.jl.  see also https://github.com/JuliaGraphics/Gtk.jl/pull/240

micahjsmith:
@bjarthur I checked out your pull request. If I add those two dependencies to deps/build.jl, then I can successfully build on my system! However, trying `using Gtk` has lots of problems. First, I had to change `gtk_version = 2` in deps/ext.jl. Then, I still get errors like this:
```
julia> using Gtk
INFO: Precompiling module Gtk...
[...]

Issue #157, opened 2015-06-24
[PkgEval] Gtk may have a testing issue on Julia 0.4 (2015-06-23)


IainNZ:
[PackageEvaluator.jl](https://github.com/IainNZ/PackageEvaluator.jl) is a script that runs nightly. It attempts to load all Julia packages and run their tests (if available) on both the stable version of Julia (0.3) and the nightly build of the unstable version (0.4). The results of this script are used to generate a [package listing](http://pkg.julialang.org/) enhanced with testing results.

#### On Julia 0.4
* On **2015-06-22** the testing status was `Tests pass.`
* On **2015-06-23** the testing status changed to `Tests fail.`

*This issue was filed because your testing status became worse. No additional issues will be filed if your package remains in this state, and no issue will be filed if it improves. If you'd like to opt-out of these status-change messages, reply to this message saying you'd like to and @IainNZ will add an exception. If you'd like to discuss [PackageEvaluator.jl](https://github.com/IainNZ/PackageEvaluator.jl) please file an issue at the repository. For example, your package may be untestable on the test machine due to a dependency - an exception can be added.*

*Test log*:
```
>>> 'Pkg.add("Gtk")' log
INFO: Cloning cache of Gtk from git://github.com/JuliaLang/Gtk.jl.git
INFO: Installing BinDeps v0.3.12
INFO: Installing Cairo v0.2.28
INFO: Installing Color v0.4.5
INFO: Installing FixedPointNumbers v0.0.7
INFO: Installing Graphics v0.1.0
INFO: Installing Gtk v0.8.2
INFO: Installing SHA v0.0.4
INFO: Installing URIParser v0.0.5
INFO: Building Cairo
WARNING: beginswith is deprecated, use startswith instead.
 in depwarn at ./deprecated.jl:62
 in beginswith at deprecated.jl:30
 in available_versions at /home/vagrant/.julia/v0.4/BinDeps/src/dependencies.jl:116
 in package_available at /home/vagrant/.julia/v0.4/BinDeps/src/dependencies.jl:111
 in can_provide at /home/vagrant/.julia/v0.4/BinDeps/src/dependencies.jl:608
 in _find_library at /home/vagrant/.julia/v0.4/BinDeps/src/dependencies.jl:451
 in allf at /home/vagrant/.julia/v0.4/BinDeps/src/dependencies.jl:622
[...]

PR #242, opened 2016-06-28
better support for different versions of gtk

bjarthur:
this PR adds aliases for libgtk2 and libgdk2, which partially fixes https://github.com/JuliaGraphics/Gtk.jl/issues/194

### DIFF
--- a/src/selectors.jl
+++ b/src/selectors.jl
@@ -41,7 +41,7 @@ if libgtk_version >= v"3"     ### should work with v >= 2.4, but there is a bug 
     end
 
     const SingleComma = r"(?<!,),(?!,)"
-    function GtkFileFilterLeaf(; name::Union{String,Void} = nothing, pattern::String = "", mimetype::String = "")
+    function GtkFileFilterLeaf(; name::Union{AbstractString,Void} = nothing, pattern::AbstractString = "", mimetype::AbstractString = "")
         filt = GtkFileFilterLeaf(ccall((:gtk_file_filter_new,libgtk), Ptr{GObject}, ()))
         if !isempty(pattern)
             name == nothing && (name = pattern)
@@ -61,7 +61,7 @@ if libgtk_version >= v"3"     ### should work with v >= 2.4, but there is a bug 
         ccall((:gtk_file_filter_set_name,libgtk), Void, (Ptr{GObject}, Ptr{UInt8}), filt, name === nothing || isempty(name) ? C_NULL : name)
         filt
     end
-    GtkFileFilterLeaf(pattern::String; name::Union{String,Void} = nothing) = GtkFileFilterLeaf(; name=name, pattern=pattern)
+    GtkFileFilterLeaf(pattern::AbstractString; name::Union{AbstractString,Void} = nothing) = GtkFileFilterLeaf(; name=name, pattern=pattern)
 
     GtkFileFilterLeaf(filter::GtkFileFilter) = filter
 
--- a/src/selectors.jl
+++ b/src/selectors.jl
@@ -1,122 +1,124 @@
-#https://developer.gnome.org/gtk2/stable/SelectorWidgets.html
+if libgtk_version >= v"2.4"
 
-#GtkColorButton — A button to launch a color selection dialog
-#GtkColorSelectionDialog — A standard dialog box for selecting a color
-#GtkColorSelection — A widget used to select a color
-#GtkHSV — A 'color wheel' widget
-#GtkFileChooser — File chooser interface used by GtkFileChooserWidget and GtkFileChooserDialog
-#GtkFileChooserButton — A button to launch a file selection dialog
-#GtkFileChooserDialog — A file chooser dialog, suitable for "File/Open" or "File/Save" commands
-#GtkFileChooserWidget — File chooser widget that can be embedded in other widgets
-#GtkFileFilter — A filter for selecting a file subset
-#GtkFontButton — A button to launch a font selection dialog
-#GtkFontSelection — A widget for selecting fonts
-#GtkFontSelectionDialog — A dialog box for selecting fonts
-#GtkInputDialog — Configure devices for the XInput extension
+    #https://developer.gnome.org/gtk2/stable/SelectorWidgets.html
 
-function push!(widget::GtkDialog, text::AbstractString, response::Integer)
-    ccall((:gtk_dialog_add_button,libgtk), Ptr{GObject},
-          (Ptr{GObject},Ptr{UInt8},Cint), widget, text, response)
-    widget
-end
+    #GtkColorButton — A button to launch a color selection dialog
+    #GtkColorSelectionDialog — A standard dialog box for selecting a color
+    #GtkColorSelection — A widget used to select a color
+    #GtkHSV — A 'color wheel' widget
+    #GtkFileChooser — File chooser interface used by GtkFileChooserWidget and GtkFileChooserDialog
+    #GtkFileChooserButton — A button to launch a file selection dialog
+    #GtkFileChooserDialog — A file chooser dialog, suitable for "File/Open" or "File/Save" commands
+    #GtkFileChooserWidget — File chooser widget that can be embedded in other widgets
+    #GtkFileFilter — A filter for selecting a file subset
+    #GtkFontButton — A button to launch a font selection dialog
+    #GtkFontSelection — A widget for selecting fonts
+    #GtkFontSelectionDialog — A dialog box for selecting fonts
+    #GtkInputDialog — Configure devices for the XInput extension
 
-#if VERSION >= v"0.4-"
-#GtkFileChooserDialogLeaf(title::AbstractString, parent::GtkContainer, action::Integer, button_text_response::=>...; kwargs...) =
-#    GtkFileChooserDialogLeaf(title::AbstractString, parent, action, button_text_response; kwargs...)
-#end
-function GtkFileChooserDialogLeaf(title::AbstractString, parent::GtkContainer, action::Integer, button_text_response; kwargs...)
-    w = GtkFileChooserDialogLeaf(ccall((:gtk_file_chooser_dialog_new,libgtk), Ptr{GObject},
-                (Ptr{UInt8},Ptr{GObject},Cint,Ptr{Void}),
-                title, parent, action, C_NULL); kwargs...)
-    for (k,v) in button_text_response
-        push!(w, k, v)
+    function push!(widget::GtkDialog, text::AbstractString, response::Integer)
+        ccall((:gtk_dialog_add_button,libgtk), Ptr{GObject},
+              (Ptr{GObject},Ptr{UInt8},Cint), widget, text, response)
+        widget
     end
-    w
-end
 
-run(widget::GtkDialog) = GLib.g_sigatom() do
-    ccall((:gtk_dialog_run,libgtk), Cint, (Ptr{GObject},), widget)
-end
-
-const SingleComma = r"(?<!,),(?!,)"
-function GtkFileFilterLeaf(; name::Union{AbstractString,Void} = nothing, pattern::AbstractString = "", mimetype::AbstractString = "")
-    filt = GtkFileFilterLeaf(ccall((:gtk_file_filter_new,libgtk), Ptr{GObject}, ()))
-    if !isempty(pattern)
-        pattern = String(pattern)
-        name == nothing && (name = pattern)
-        for p in split(pattern, SingleComma)
-            p = replace(p, ",,", ",")   # escape sequence for , is ,,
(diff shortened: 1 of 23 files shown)


vtjnash:
`gtk_version` isn't properly responsive to this, but it's also not really necessary to get the build script to work (Gtk.jl doesn't care how the deps got installed as long as they can be loaded).

bjarthur:
`gtk_version` is defined in `ext.jl`, which is not included in `build.jl`.  i suggest we move the definition to the latter, and then in `src/` change all references to `gtk_version` to `libgtk_version`.  does that make sense?

bjarthur:
it seems that all of `ext.jl` is just a hack for https://github.com/JuliaLang/BinDeps.jl/issues/146.  i don't think we can move the definition of `gtk_version` to `build.jl` until that is fixed.

in the meantime, in addition to adding gtk2 aliases for `BinDeps`, i've changed all references to `gtk_version` in `src` to `libgtk_version`, as well as fixed `tests` to work with gtk2.

i think this is ready to merge, unless the workaround for an empty `deps.jl` is no longer necessary, in which case i'd be happy to back out the hack.

bjarthur:
i think this is ready to merge.  the tests only fail because travis uses gtk2 and gtk_version is hard coded to 3.  could someone please review please?  thanks.

hustf:
I tried to test this but got some error messages. I suppose we'll find why in the status of my other packages. But since I don't often do this, I put in the procedure for my own reference first. 

```
Julia: Pkg.checkout("Gtk")
Git bash: 
C:\Users\F\.julia\v0.4\Gtk>git fetch origin pull/242/head:TESTPR242
remote: Counting objects: 30, done.
remote: Compressing objects: 100% (10/10), done.
remote: Total 30 (delta 22), reused 19 (delta 19), pack-reused 1
Unpacking objects: 100% (30/30), done.
From https://github.com/JuliaGraphics/Gtk.jl
 * [new ref]         refs/pull/242/head -> TESTPR242

C:\Users\F\.julia\v0.4\Gtk>git checkout TESTPR242
Switched to branch 'TESTPR242'
```
Inspection of files now show the PR changes locally.

hustf:
```
julia> Pkg.test("Gtk")
INFO: Testing Gtk
INFO: Recompiling stale cache file C:\Users\F\.julia\lib\v0.4\Gtk.ji for module Gtk.
INFO: Recompiling stale cache file C:\Users\F\.julia\lib\v0.4\FixedPointNumbers.ji for module FixedPointNumbers.
INFO: Recompiling stale cache file C:\Users\F\.julia\lib\v0.4\Graphics.ji for module Graphics.
INFO: Recompiling stale cache file C:\Users\F\.julia\lib\v0.4\Colors.ji for module Colors.
INFO: Recompiling stale cache file C:\Users\F\.julia\lib\v0.4\ColorTypes.ji for module ColorTypes.
INFO: Recompiling stale cache file C:\Users\F\.julia\lib\v0.4\Cairo.ji for module Cairo.

(julia.exe:1488): GModule-CRITICAL **: g_module_symbol: assertion 'symbol != NULL' failed
(repeats about 40 times)
The following messages:
   WARNING: Error in @guarded callback
are expected and a sign of normal operation.
WARNING: Error in @guarded callback
ERROR: UndefVarError: k not defined
WARNING: Error in @guarded callback
ERROR: UndefVarError: k not defined
WARNING: Error in @guarded callback
ERROR: UndefVarError: k not defined
WARNING: Error in @guarded callback
ERROR: UndefVarError: k not defined

(julia.exe:1488): GModule-CRITICAL **: g_module_symbol: assertion 'symbol != NULL' failed
INFO: Gtk tests passed
```

hustf:
I'm on Windows 10 64 bit, Julia 0.4.6. Status of Pkg requirements:
 
- Gtk                           0.10.1+            TESTPR242
 - BinDeps                       0.4.0
 - SHA                           0.1.2
 - URIParser                     0.1.5
 - Cairo                         0.2.33
 - Graphics                      0.1.3
 - Colors                        0.6.6
 - Reexport                      0.0.3
 - FixedPointNumbers             0.1.4
 - ColorTypes                    0.2.5
( - Reactive                      0.3.3+             master)

I am not sure if Reactive is a requirement.

I'm keeping the PR checked out in case you have anything you'd like me to try.

bjarthur:
the uncoolness should be fixed now

hustf:
I do not understand what you mean? I Pkg.unpinned, Pkg.updated and repeated the process above with a new test branch.
The repeating error message number has changed from :1488 to: 

(julia.exe:5828): GModule-CRITICAL **: g_module_symbol: assertion 'symbol != NULL' failed

How cool is that?

bjarthur:
works for me on scientific linux 6.5 and gtk 2.24.23.  perhaps it's a windows problem?  or, what version of gtk you are using?  the latter can be determined by `using Gtk; Gtk.libgtk_version`.

either way, to isolate the problem you'll probably need to run the tests manually, copying and pasting each section in by hand until the problem arises.  look in ~/.julia/v0.4/Gtk/tests/runtests.jl.

bjarthur:
note that the `@FileChooserDialog` near line 504 of `test/gui.jl` cause seg faults on my linux / gtk2 system.  i think this is related to https://github.com/JuliaGraphics/Gtk.jl/issues/161.  strangely though, this test passes on my os x / gtk3 system.  not sure what to do here.

tknopp:
@bjarthur: Might I ask, if gtk2 is the only option for you? I am not sure if it is feasible in the long run to support different Gtk versions in this package.

bjarthur:
@tknopp it's currently the only option for some of my clients.  would be nice to keep v2 support for at least a little while longer.  the only thing i'm aware of that doesn't work is the FileChooserDialogs.

tknopp:
So you are using v2 in a productive environment?
I am not in principle against v2 support. But currently I would be quite defensive to which degree v2 is actually supported. A non-working FileChooserDialog is at least a pretty important thing not working.

hustf:
Bjarthur, sorry for the later reply. I am using v"3.18.2".

hustf:
Bjarthur, I also tried on another Win10 computer. I tried various approaches: Pkg.test.. , Pkg.build.., using Gtk. All fails. For example:
```
julia> using Gtk;Gtk.libgtk_version
INFO: Precompiling module Gtk...
ERROR: LoadError: LoadError: error compiling anonymous: could not load library "libgobject-2.0-0"
The specified module could not be found.
```
Note that I get similar issues on this computer both on the tagged version and on Master.

bjarthur:
@hustf  if you're using gtk3 and having problems, then you should open a new issue.  this issue is about gtk2 support.

bjarthur:
@tknopp yes, i'm using gtk2 in a production environment.  at least for the next few months.

i don't think we need to officially support gtk2.  a "use at your own risk" warning is fine with me.  but if someone using gtk2 fixes a bug and submits a PR, then great, we should take it.  i don't personally need the FileChooserDialog so will leave that fix to someone that does.

as for this PR, it mostly just replaces `if gtk_version == 3` with e.g. `libgtk_version >= v"3.18.0"` and adds similar logic to disable those features that aren't present in gtk2.  such logic is useful going forward as gtk3 adds new features and eventually gtk4 is released.  much better to query the library of it's version, which include the minor and point release numbers, instead of a user defined variable which only provides the major number.  all this PR does is backport that logic to gtk2, and make consistent use of it for gtk3.

tknopp:
Yes this seems like a good approach to me.
However, CI seem to currently fail and I would several unrelated changes. For instance you removed `empty!` that has been recently added. Was this intentional?

bjarthur:
tests are failing on all recent os x PRs b/c of https://github.com/Homebrew/homebrew-core/pull/3279

bjarthur:
tests are now passing on OS X.  still failing on linux, presumably because travis uses gtk2, as they pass on my local machine when i manually set `gtk_version` to 2.

please note that this PR now consists of 5 commits, which overall add better support for gtk2.  i could split them into separate PRs if desired:

1. systematically use libgtk_version instead of gtk_version throughout.  this sets the stage for eventually getting rid of the manually set `gtk_version`.

2. add the gtk2 libs to the names that BinDeps searches for

3. sleep for longer when min/max'imizing windows so that tests pass on slow systems.  this change was made once before, and then accidentally (i think) reverted.  i reinstate that change here as otherwise tests fail on my old machine.

4. update the tests to use `@testset` so it's easier to debug what doesn't work.  for now i haven't indented the actual tests, so that it's easier to see what i've changed in the diff.  once reviewed, i could go back an add four spaces.

5. update `@linux_only` to `is_linux` etc. for julia v0.5 compatability

tknopp:
@vtjnash @timholy:

I am not really capable of reviewing this PR. You? My general feeling I have expressed above is that we should drop gtk2 support since `Gtk.jl` cannot be to broad given that there is also a certain lack of manpower on this project. The python bindings also do not support gtk2.

bjarthur:
perhaps it's better to describe this PR as providing better support for different versions of gtk, not just 2 vs 3, but different point releases of 3 as well.  the only commit specific to gtk2 is the one that adds the gtk2 lib names to the BinDeps search.

bjarthur:
also, can anyone confirm what version of gtk the travis boxe uses?  the linux tests pass on my gtk2 box when i have gtk_version manually set to 2 in deps/ext.jl, and also pass on by gtk3 box when it remains set to the default 3.  but they fail on travis, so i assume it is gtk2.

tknopp:
no travis uses gtk3 as one can see in the build logs

bjarthur:
@timholy `{open,save}_dialog` / `GtkFileChooserDialog` is broken.  it has been so for awhile i believe, and only came to light when i refactored the tests to use `@testset`.  i'm at a loss as to how to fix it.  could you please take a look?

the error returned is a `MethodError`:

```
julia> using Gtk.ShortNames, Gtk.GConstants

julia> @FileChooserDialog("Select file", @Null(), GtkFileChooserAction.OPEN,
                                   (("_Cancel", GtkResponseType.CANCEL),
                                    ("_Open", GtkResponseType.ACCEPT)))
ERROR: MethodError(Gtk.GtkFileChooserDialogLeaf,("Select file",Gtk.GtkNullContainer(<NULL>),0,(("_Cancel",-6),("_Open",-3))))
 in #GtkFileChooserDialogLeaf#90(::Array{Any,1}, ::Type{T}, ::String, ::Vararg{Any,N}) at /home/arthurb/.julia/v0.5/Gtk/src/GLib/gtype.jl:212
 in Gtk.GtkFileChooserDialogLeaf(::String, ::Gtk.GtkNullContainer, ::Int32, ::Tuple{Tuple{String,Int32},Tuple{String,Int32}}) at /home/arthurb/.julia/v0.5/Gtk/src/GLib/gtype.jl:211
```

@Matt5sean3 's solution to wrap strings in `Gtk.String` didn't work.

thanks.

bjarthur:
linux tests work now.  problem was that the travis box seemingly has both gtk2 and gtk3 installed, and when BinDeps was configured to look for both, it returned gtk2.  so i have instead added comments to `deps/{build,ext}.jl` as to what changes to make for gtk2 if desired.

perhaps when [this](https://github.com/JuliaLang/BinDeps.jl/issues/146) and [this](https://github.com/JuliaGraphics/Gtk.jl/issues/157#issuecomment-114687693) are fixed we could have `Gtk.jl` auto-detect what version is installed.  but by that time gtk2 will probably have long since died.

so the two main things this PR adds are:

1. a systematic use of the library provided `libgtk_version`, instead of the user provided `gtk_version`, to enable / disable certain features, and

2. a refactoring of the tests to use `@testset`, which revealed a couple otherwise unknown failures.

pending a review, and a fix for `FileChooserDialog`, i think it's ready to go.

timholy:
https://github.com/JuliaGraphics/Gtk.jl/pull/242#issuecomment-245643520 works for me locally. Still a problem for you?

bjarthur:
oi, my bad, too hasty of a rebase on master.  all tests are green now.  thanks.

src/selectors.jl
```diff
@@ -1,122 +1,124 @@
-#https://developer.gnome.org/gtk2/stable/SelectorWidgets.html
+if libgtk_version >= v"3"     ### should work with v >= 2.4, but there is a bug for v < 3
```
timholy: Just checking, is this comment still relevant?

timholy:
LGTM. I'll merge soon if there are no objections.

src/selectors.jl
bjarthur: well, nominally FileChooserDialog was added to gtk in v2.4 according to their docs.  but i can't get it to work with the julia bindings using gtk2.  so i put a disclaimer there in case someone needed it badly enough that they wanted to fix it.

timholy:
Thanks!

PR #244, opened 2016-07-10
place cursor, GdkRGBA that works with tag, expanded constructor of di…

guidocalvano:
I couldn't merge the RGBA and GdkRGBA because the rgba values were of different primitive types. I also couldn't figure out how to expose GdkRGBA analogous to for instance @TextBuffer, so now it can be accessed with Gtk.GdkRGBA.

### DIFF
--- a/src/displays.jl
+++ b/src/displays.jl
@@ -16,7 +16,7 @@ convert{U<:Unsigned}(::Type{U},x::RGB) = convert(U,(x.r)|(x.g>>8)|(x.b>>16))
 
 immutable RGBA
     r::UInt8; g::UInt8; b::UInt8; a::UInt8
-    RGBA(r,g,b) = new(r,g,b)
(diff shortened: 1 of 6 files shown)


src/displays.jl
```diff
@@ -16,7 +16,7 @@ convert{U<:Unsigned}(::Type{U},x::RGB) = convert(U,(x.r)|(x.g>>8)|(x.b>>16))
 
 immutable RGBA
     r::UInt8; g::UInt8; b::UInt8; a::UInt8
-    RGBA(r,g,b) = new(r,g,b)
+    RGBA(r,g,b,a) = new(r,g,b,q)
```
vtjnash: `r,g,b,a`

vtjnash:
Thanks, this looks right to me. I didn't realize that they were exposing different primitive types. Can you get CI passing and then I can merge this.

PR #245, opened 2016-07-21
fix broken syntax on 0.5

KristofferC:
looks correct @stevengj?

ref https://github.com/JuliaLang/julia/pull/17393#issuecomment-234122725

### DIFF
--- a/src/container.jl
+++ b/src/container.jl
@@ -18,7 +18,7 @@ function append!(w::GtkContainer, children)
     end
     w
 end
-Base.|>(parent::GtkContainer, child::Union{GObject,AbstractString}) = push!(parent, child)
+@compat Base.:|>(parent::GtkContainer, child::Union{GObject,AbstractString}) = push!(parent, child)
 
 start(w::GtkContainer) = glist_iter(ccall((:gtk_container_get_children,libgtk), Ptr{_GList{GObject}}, (Ptr{GObject},), w))
 next(w::GtkContainer, list) = next(list[1],list)


stevengj:
Yep.

tknopp:
CI seems to be unhappy.

stevengj:
Seems like an unrelated failure?

KristofferC:
Only on Mac. Seems to be unrelated?

lobingera:
> Only on Mac.

Some Mac users have strong oppinions on that...

I'm not an expert on this, but could it be another Homebrew hiccup?
```
> (<unknown>:50507): GdkPixbuf-WARNING **: Cannot open pixbuf loader module file '/Users/travis/.julia/v0.4/Homebrew/deps/usr/lib/gdk-pixbuf-2.0/2.10.0/loaders.cache': No such file or directory

This likely means that your installation is broken.

Try running the command

  gdk-pixbuf-query-loaders > /Users/travis/.julia/v0.4/Homebrew/deps/usr/lib/gdk-pixbuf-2.0/2.10.0/loaders.cache

to make things work again for the time being.
```

tknopp:
@tkelman has activated CI on mac in #213.
What should we for this PR? Ignore CI on Mac?

tkelman:
it worked a few weeks ago, ref https://travis-ci.org/JuliaGraphics/Gtk.jl/builds/142129851 - but Homebrew.jl has gone through some overhauls recently

tknopp:
@tkelman: So, is it ok to ignore this for the current PR? The fix looks ok.

KristofferC:
@lobingera my point was that if it only fails on mac then it is likely not the PR's fault since the change should be OS agnostic.

lobingera:
@KristofferC I agree. But if building gtk3+ on travis creates a problem, another issue should be opened.

KristofferC:
Please do.

Issue #247, opened 2016-08-01
GtkFileFilter construction fails in julia 0.4


Matt5sean3:
Within julia 0.4 the `GtkFileFilter` fails due to a MethodError. Within julia 0.4 the empty string literal, `""`, is an `ASCIIString` but `Gtk.String`, which seems to be a shim for the string changes in julia 0.5, is a `UTF8String`. The default values utilize the empty string literal which causes type checking to fail.

This issue arises in both `v0.10.1` and the `master` branch.

Matt5sean3:
Switching from `String` to `AbstractString` seems to fix the problem with minimal issues and doesn't reintroduce warnings in julia 0.5 from what I'm seeing. If the concrete type needs to be ensured the arguments could be passed through a conversion to make sure the type is correct.

I'll probably create a pull request with this fix later today.

Matt5sean3:
Anyone else experiencing this can use this work-around until the problem is solved. Pattern and mimetype must be set explicitly, their default arguments cause failure. Using just a string for a filter in `open_dialog` will still fail even if it's a UTF8String because the mimetype default argument will still cause a `MethodError`

```julia
filter = @GtkFileFilter(pattern=UTF8String("*.csv"), mimetype=UTF8String("text/csv"))
file_path = open_dialog("Open a File", GtkNullContainer(), (filter, ))
```

tknopp:
I can confirm this issue. Question is a little how this will work in julia 0.5 where `String` is again a concrete type.

Matt5sean3:
`AbstractString` still exists in 0.5, or at least in 0.5-rc3. From there, force conversion to the `String` type to guarantee compatibility with C. The `String` type is defined locally in the Gtk package, via compat I think, so this is a conversion to a UTF8String in 0.4, which I'm not sure how well C deals with. This is how I solved it in my pull request at least.

Matt5sean3:
The fix has been merged in with test cases to prevent the problem from reoccurring so I'll close the issue.

PR #248, opened 2016-08-01
Fixed GtkFileFilter MethodError for ASCIIStrings in julia 0.4

Matt5sean3:
This pull request is meant to fix the problem described in #247 

In the current version
```julia
@GtkFileFilter("*.txt")
```
will fail with a `MethodError` in julia 0.4.6

This also means that 
```julia
open_dialog("Open a File", GtkNullContainer(), ("*.txt",  ))
```
fails in julia 0.4 while using that constructor.

Until this problem is fixed file filtering in dialogs is impossible in julia 0.4

### DIFF
--- a/src/selectors.jl
+++ b/src/selectors.jl
@@ -39,15 +39,17 @@ run(widget::GtkDialog) = GLib.g_sigatom() do
 end
 
 const SingleComma = r"(?<!,),(?!,)"
-function GtkFileFilterLeaf(; name::Union{String,Void} = nothing, pattern::String = "", mimetype::String = "")
+function GtkFileFilterLeaf(; name::Union{AbstractString,Void} = nothing, pattern::AbstractString = "", mimetype::AbstractString = "")
     filt = GtkFileFilterLeaf(ccall((:gtk_file_filter_new,libgtk), Ptr{GObject}, ()))
     if !isempty(pattern)
+        pattern = String(pattern)
         name == nothing && (name = pattern)
         for p in split(pattern, SingleComma)
             p = replace(p, ",,", ",")   # escape sequence for , is ,,
             ccall((:gtk_file_filter_add_pattern,libgtk), Void, (Ptr{GObject}, Ptr{UInt8}), filt, p)
         end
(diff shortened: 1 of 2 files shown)


Matt5sean3:
After further examination, file filtering isn't strictly impossible, just a bit unwieldy and has some unexpected requirements:

This will succeed in creating an opener for csv files.
```julia
filter = @GtkFileFilter(; pattern=UTF8String("*.csv"), mimetype=UTF8String("text/csv"))
file_path = open_dialog("Open a File", GtkNullContainer(), (filter, ))
```

tknopp:
@Matt5sean3: Seems correct to me. Could you please add a test for `@GtkFileFilter` so that we can check in the future that it works?

Thanks!

Matt5sean3:
I've added tests for the current behavior seen in the `@GtkFileFilter` constructor. They seem to pass under Linux at least for v0.4 and v0.5.

tknopp:
Thanks for the fix and adding the tests. I merged it for now, but why is the `GtkFileFilterInfo` just defined in the test file? Isn't this part of the API?

Matt5sean3:
I added `GtkFileFilterInfo` just as part of the tests because the only places I saw it used was for `gtk_file_filter_filter`, and in the definition of `GtkFileFilterFunc` which is used by `gtk_file_filter_add_custom`, neither of which are wrapped anywhere in the current Julia Gtk package. This would have left that type without a use within the Gtk package itself. Adding wrappers for the functions that use `GtkFileFilterInfo` to the main package would have expanded the scope of the pull request beyond the stated goal so I kept it limited to the test files.

Matt5sean3:
Also, there are some garbage collection pitfalls that could arise with the `GtkFileFilterInfo` type as currently defined, as the strings pointed to by the `filename`, `uri`, `display_name`, and `mime_type` members could be garbage collected if they aren't referenced elsewhere making the `GtkFileFilterInfo` as currently defined a bit too unsafe for general usage in my estimation.

tknopp:
Yea thats seem reasonable then.

PR #253, opened 2016-08-22
Fix super->supertype deprecation warning

timholy:
I'm going to tag a new release, since the current one fails on 0.5 and makes packages like ProfileView fail their tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,12 @@
 language: julia
 sudo: false
-matrix:
-  include:
-    - os: linux
-      julia: 0.4
-      env: TESTCMD="xvfb-run julia"
-    - os: osx
-      julia: 0.4
-      env: TESTCMD="julia"
+os:
+  - linux
+  - osx
+julia:
+  - 0.4
+  - 0.5
+  - nightly
 notifications:
     email: false
 addons:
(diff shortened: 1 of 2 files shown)


lobingera:
how can deprecations warnings break testing?

timholy:
They don't. I just wanted to fix test warnings before tagging.

timholy:
The last tag of Gtk was in mid-June.

src/GLib/glist.jl
```diff
@@ -16,9 +16,9 @@ immutable _GList{T} <: _LList{T}
     prev::Ptr{_GList{T}}
 end
 eltype{T}(::Type{_LList{T}}) = T
-eltype{L<:_LList}(::Type{L}) = eltype(super(L))
+eltype{L<:_LList}(::Type{L}) = eltype(supertype(L))
```
tkelman: is this in Compat 0.8.0 ?

.travis.yml
```diff
@@ -3,10 +3,16 @@ sudo: false
 matrix:
   include:
     - os: linux
-      julia: 0.4
+      julia:
+        - 0.4
+        - 0.5
+        - nightly
```
tkelman: I don't think it likes things being nested. Probably better to just set TESTCMD based on the output of uname

timholy:
There's some Homebrew problem. Will this be exacerbated if I tag a new version, or is it already an issue now?

tkelman:
ref https://github.com/JuliaLang/Homebrew.jl/issues/121 so it's been a problem for a while, but may be work-around-able with a `Homebrew.postinstall` call? cc @staticfloat so this doesn't fall off your radar completely

timholy:
OK, so I won't let this hold up tagging.

PR #257, opened 2016-09-25
change bool to Bool and add a test for visible

jonathanBieler:
`visible` wasn't working with 0.5, it was an issue with `bool`. I changed all of them to `Bool`and added a test for `visible`.


tknopp:
Thanks!

PR #259, opened 2016-10-13
Fix syntax error in typealias LListPair

gaika:
with v0.6:
 using Gtk
INFO: Precompiling module Gtk.
ERROR: LoadError: LoadError: LoadError: syntax: unexpected end

### DIFF
--- a/src/GLib/glist.jl
+++ b/src/GLib/glist.jl
@@ -58,7 +58,7 @@ end
 if VERSION >= v"0.4-"
     typealias LListPair{L} Tuple{LList, Ptr{L}}
 else
-    typealias LListPair{L}(LList, Ptr{L})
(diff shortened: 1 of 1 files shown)


tkelman:
what version of julia are you using? on 0.6-dev the parsing of typealias changed recently but not sure if it should have broken this

gaika:
versioninfo()
Julia Version 0.6.0-dev.950
Commit ec80011 (2016-10-12 15:47 UTC)

vtjnash:
@tkelman yes, the current syntax I believe was working around a different parser bug. I'll fix this is my PR.

tkelman:
can't you delete the else now that julia 0.4 is the minimum supported?

vtjnash:
probably. I would prefer doing that all at once though

PR #260, opened 2016-10-14
updates and fixes

vtjnash:
style improvements, a few typos, and preparation for https://github.com/JuliaLang/julia/pull/18754

### DIFF
--- a/test/style_test.css
+++ b/test/style_test.css
@@ -1,4 +1,4 @@
 GtkLabel {
   color: blue;
-  font: 16px;
+  font: 16px serif;
 }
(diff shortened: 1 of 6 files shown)


test/style_test.css
```diff
@@ -1,4 +1,4 @@
 GtkLabel {
   color: blue;
-  font: 16px;
+  font-size: 16px;
```
fcard: Travis' gtk can't parse `font-size` for some reason, which is why I used just `font`. (https://github.com/JuliaGraphics/Gtk.jl/pull/204#issuecomment-160483195) I did forget to add the font-family value, sorry :P

vtjnash: sigh, thanks for pointing that out. I'll switch back to `font`

tknopp:
@vtjnash merge? or is this WIP?

PR #282, opened 2017-01-17
whitespace

vtjnash:
no functional changes, just making the whitespace formatting more consistent


tknopp:
lgtm. Not sure about the whitespace in type parameters though. Here for instance it is omitted:

https://github.com/JuliaLang/julia/blob/master/base/sparse/sparsematrix.jl#L8

vtjnash:
Yeah, the formatting of base isn't very clean or consistent.

Issue #263, opened 2016-10-20
Maintenance of Gtk.jl 


tknopp:
I am opening this issue to discuss the maintenance of Gtk.jl. It seems that currently Gtk.jl has more or less no defined maintainer. Is this correct? @vtjnash has driven this project (thanks so much!!!) but is currently more involved in core Julia infrastructure if I am correct.

I think if this is the case we could communicate this somewhere so that there is no direct expectation from newcomers that issues are directly solved. Maybe even "help for maintenance" message could be communicated.

I hope this is ok to discuss this in an open issue?!

Pinging some people that contributed to this package: @vtjnash @timholy @Matt5sean3 @bjarthur @lobingera

bjarthur:
who currently has push access to Gtk.jl?  is this info publicly available through the github UI?

tknopp:
don't know, this seems a listing:

https://github.com/JuliaGraphics/Gtk.jl/network/members

but I do not know if this is publicity available (I am currently logged in).

lobingera:
i think only the owners have push access and only the owners of JuliaGraphics. I'm listed as contributor, but i'm sure my PR was merged by someone else and also my work on Cairo was/is pulled by someone with write access.

tknopp:
That seems correct but I opened this issue not so much about the current status of people with commit access (all of JuliaGraphics have) but more about discussing future development. I use Gtk.jl intensively (as a user) and am pretty satisfied with it. @timholy seems to be also a power user of this package.

tknopp:
I think people that pushed things to master in the past were primary @vtjnash, @timholy, and I have also pushed some recent things.

timholy:
I do have push access; I am likely to get back to this package in the next month or so once I get serious about the ImageView rewrite, but I certainly don't have time to act as a maintainer now. Perhaps we need to increase the pool of folks with commit access?

tknopp:
Hey Tim, yes, this thread is also meant as some kind of "call for maintainer/contributor" message.

timholy:
@tknopp, sounds like you may be volunteering for the job? :smile:

tknopp:
:-), I feared that you ask that. Partially I can do that. But there are areas within this package that I don't fully understand. And time I can spend on this is limited.

timholy:
It would be great to have you do whatever you can!

timholy:
You've been doing a great job, @tknopp---I can't thank you enough.

Any thoughts on tagging a new release?

[...]

PR #285, opened 2017-02-22
Use isdefined and Compat instead of unspecific VERSION conditions

tkelman:
take! requires minimum version of 0.9.5

ref https://github.com/JuliaGraphics/Gtk.jl/issues/263#issuecomment-281695222

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -4,5 +4,5 @@ Graphics 0.1
 BinDeps
 @windows WinRPM
 @osx Homebrew
-Compat 0.8.0
+Compat 0.9.5
 BaseTestNext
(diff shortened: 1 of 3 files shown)


vtjnash:
Gtk is not compatible with Compat.String. It could use GLib.String if this really bothers you

tkelman:
it's only used in the two places that call `take!` here

timholy:
@tknopp, I turned on [attobot](https://github.com/attobot/attobot) so now you can make new releases using the GitHub GUI.

timholy:
And I submitted a PR for v0.11.0

tknopp:
@timholy thanks, attobot is really great!

Issue #286, opened 2017-03-01
Fails to precompile on julia 0.6


CarloLucibello:
I hit two errors when trying to import Gtk.jl on julia 0.6. The first was given by
```julia
immutable GtkTextRange <: Range
....
```
It appears that Range is not an abstract type anymore. I did the most brutal thing and changed the definition to 
```julia
immutable GtkTextRange
....
```
and it seemed to work out.

The second error is 
```julia
ERROR: LoadError: LoadError: ArgumentError: method definition for getindex at /home/carlo/.julia/v0.6/Gtk/src/displays.jl:256 has free type variables
Stacktrace:
 [1] include_from_node1(::String) at ./loading.jl:539
 [2] include(::String) at ./sysimg.jl:14
 [3] include_from_node1(::String) at ./loading.jl:539
 [4] include(::String) at ./sysimg.jl:14
 [5] anonymous at ./<missing>:2
while loading /home/carlo/.julia/v0.6/Gtk/src/displays.jl, in expression starting on line 256
while loading /home/carlo/.julia/v0.6/Gtk/src/Gtk.jl, in expression starting on line 65
```
caused by the definition
```julia
typealias Index Union{Integer, AbstractVector{TypeVar(:I, Integer)}}
```
I changed the line to
```julia
const Index = Union{Integer, AbstractVector{T} where T<:Integer}
```
and the error was gone. I don't know of any fix which is both 0.5 and 0.6 compatible though. Could we restrict to Vector{Int}?

With these two fixes a ProfileView window is displayed correctly, I only get tons of warnings.

Bye,
Carlo

smldis:
8/8 tests passed with julia master, Gtk.jl master and my pull request

tknopp:
yes but it does not work anymore for 0.4 and 0.5

timholy:
@smldis, thanks for stepping forward to improve Gtk! But see #287, which (if merged soon) will fix these and other errors.

smldis:
Thank you all!
Issue solved in master.

PR #287, opened 2017-03-01
drop julia 0.4 + fix errors and deprecations in 0.6

CarloLucibello:
I'm leaving inline comments on a few changes I'm not sure about

PR #290, opened 2017-03-04
Fix two building errors by appling Tim Holy suggestions

smldis:
Fix two building errors by appling Tim Holy suggestions: https://github.com/JuliaGraphics/Gtk.jl/issues/286


timholy:
Fixed in #287.

PR #291, opened 2017-03-07
defined iteratorsize(::TreeIterator) instead of an O(n) length method

tkelman:


tknopp:
is `length` not necessary for the iterator protocol?

tkelman:
No, the iterator protocol only needs start, next, done.

map and collect need either length or this trait (the default should arguably be SizeUnknown)

PR #310, opened 2017-05-05
RCF: better shortcut facilities

jonathanBieler:
This solves a couple of issues:

- It makes it a bit easier to create keyboard shortcuts and to check them against events.
- It checks correctly shortcuts, taking into account CapsLock and such.
- It introduce platform dependent Modifiers (PrimaryModifier and SecondaryModifer) so you can have consistent shortcuts between windows/linux and OS X (PrimaryModifier refers to Control on windows and Command on OS X).
- It checks correctly (I think) the Command key on OS X (which is also the Meta key by default).

### DIFF
--- a/src/Gtk.jl
+++ b/src/Gtk.jl
@@ -90,6 +90,7 @@ using .GConstants
 
 include("windows.jl")
 include("gl_area.jl")
+include("shortcuts.jl")
 
 # Alternative Interface (`using Gtk.ShortNames`)
 module ShortNames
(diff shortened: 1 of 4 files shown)


yakir12:
Too bad this hasn't been merged yet, I sure could use this now...

yakir12:
@tknopp, PM you on Discourse.

tknopp:
@jonathanBieler: should this be merged?

jonathanBieler:
I think it would be nice yes. That said I'm not super convinced by `doing(Shortcut, event)` (I had an Action type before that made more sense with `doing`) but I can't think of a better name right now (`isshortcut` isn't nice).

Issue #292, opened 2017-03-12
Icons not found


jonathanBieler:
I'm not sure what changed but some icons do not work anymore on OS X. 

```julia
b = GtkToolButton("gtk-save-as") 

win = GtkWindow(b)

showall(win)

(<unknown>:47866): Gtk-WARNING **: Error loading theme icon 'edit-find' for stock: Icon 'edit-find' not present in theme Adwaita
```

It might be because "gtk-save-as" is deprecated in Gtk, using "document-save-as" instead doesn't throw a warning but no icon is displayed instead.

tknopp:
I think there was some Homebrew.jl issue about this

tknopp:
https://github.com/JuliaGraphics/Gtk.jl/issues/180#issuecomment-282029796

tknopp:
Could you try this?

jonathanBieler:
My bad I should have searched better, the `Homebrew.add("gnome-icon-theme")` fixes it. Maybe that should be included in the build process to make sure it's installed.

tknopp:
As long as this is fixed its absolutely good to report the issue. Not sure what the proper fix is.

timholy:
Pull request? Only folks with macs can really test such PRs, and it may be the case that none of the primary developers of this package are mac users. (I'm certainly not.)

tknopp:
I can test on a mac. The issue here is, however, an Homebrew thing where I am no expert for.

jonathanBieler:
I would just insert a `Homebrew.add("gnome-icon-theme")` [here](https://github.com/jonathanBieler/Gtk.jl/blob/master/deps/build.jl#L53) but I don't really understand how BinDeps works so that might not be the correct solution.

tknopp:
Does not hurt there. So +1 for a PR ;-)

PR #314, opened 2017-06-05
Update installation.md

Evizero:
related: #292


timholy:
Maybe quote the warning message? People seem most likely to find this by googling for the warning, would probably make it much easier to find.

timholy:
Thanks!

PR #320, opened 2017-07-28
Fixes #319, also improves autogen 0.6 compatibility

### LINKED ISSUE #319: Incorrect Gtk.GConstants.GDK_KEY mapping for 'o' and 'O'
While working on creating my own easier-to-remember key bindings for Gtk, I came across what I believe is a small error in the generated values for the keys 'o' and 'O' in gconsts3. As seen [here](https://github.com/JuliaGraphics/Gtk.jl/blob/master/gen/gconsts3#L3043) and [here](https://github.com/JuliaGraphics/Gtk.jl/blob/master/gen/gconsts3#L3076), these two values do not match with the rest of their alphabetical brethren. I confirmed that the values that these indicate are different from what Gtk returns upon pressing these keys. Specifically, GDK_KEY_O should be 0x004f and GDK_KEY_o should be 0x006f.

jpsamaroo:
This is a fix for issue #319 , where autogen'd constants don't match values received from Gtk keyboard events. To be specific, GDK_KEY bindings in `gconsts3` with a value ending in `f` were missing the preceding `0` and following `f` from the binding value (i.e. 0x006f would appear as 0x06).

I initially attempted to run `gtk_auto_gen.jl` to generate the bindings, but ran into header path issues that caused the auto-generated `gconsts3` file to miss the key bindings. Therefore, a simple find-and-replace was done to correct the incorrect key bindings. I also made `gtk_auto_gen.jl` and other gen dependencies 0.6-compatible in the process. Please let me know if I should instead remove those changes and just leave the keybinding fixes.

NOTE: This is my first PR, and I probably totally messed it up. Please let me know if there's something I need to correct, or if I should scrap this and start a new PR.

### DIFF
--- a/gen/gconsts3
+++ b/gen/gconsts3
@@ -3018,13 +3018,13 @@ $(Expr(:toplevel, :(module GConstants
     const GDK_KEY_comma = 0x002c
     const GDK_KEY_minus = 0x002d
     const GDK_KEY_period = 0x002e
-    const GDK_KEY_slash = 0x02
+    const GDK_KEY_slash = 0x002f
     const GDK_KEY_colon = 0x003a
     const GDK_KEY_semicolon = 0x003b
     const GDK_KEY_less = 0x003c
     const GDK_KEY_equal = 0x003d
     const GDK_KEY_greater = 0x003e
-    const GDK_KEY_question = 0x03
+    const GDK_KEY_question = 0x003f
     const GDK_KEY_at = 0x0040
     const GDK_KEY_A = 0x0041
     const GDK_KEY_B = 0x0042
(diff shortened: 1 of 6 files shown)


tknopp:
If I understand this does not solve the root of the issue right?

jpsamaroo:
That is true, I'm not sure why the autogen stripped these characters out of the keybindings, nor am I sure that the autogen even currently works. My assumption, however, was that if these bindings don't change anytime soon (which is probably a bad assumption to make), then the bindings can be fixed in-place without having to confirm that the autogen will correctly produce them next time it is run. This is really just a band-aid for the short term.

tknopp:
Thats fine for the moment. @vtjnash might know the real fix.

PR #322, opened 2017-08-16
remove BaseTestNext from REQUIRE since it's not used anywhere

bjarthur:
am i correct in thinking this is cruft?


bjarthur:
the test failures on OS X are unrelated.  can be merged i think.  (i used to have write access i thought, but seemingly not anymore.)

tknopp:
Thanks

Issue #180, opened 2015-09-18
Gtk on OS X 10.10.5, v0.4.0-rc1, can't find icons


joshday:
Moving this to Gtk from https://github.com/JuliaGraphics/Immerse.jl/issues/10.  I am trying to run the following example
```julia
using Gtk.ShortNames
tb1 = @ToolButton("gtk-open")
toolbar = @Toolbar(tb1)
win = @Window(toolbar)
showall(win)
```
which returns:
```julia
julia> showall(win)
Gtk.GtkWindowLeaf(name="", parent, width-request=-1, height-request=-1, visible=TRUE, sensitive=TRUE, app-paintable=FALSE, can-focus=FALSE, has-focus=FALSE, is-focus=FALSE, can-default=FALSE, has-default=FALSE, receives-default=FALSE, composite-child=FALSE, style, events=0, no-show-all=FALSE, has-tooltip=FALSE, tooltip-markup=NULL, tooltip-text=NULL, window, opacity=1.000000, double-buffered, halign=GTK_ALIGN_FILL, valign=GTK_ALIGN_FILL, margin-left, margin-right, margin-start=0, margin-end=0, margin-top=0, margin-bottom=0, margin=0, hexpand=FALSE, vexpand=FALSE, hexpand-set=FALSE, vexpand-set=FALSE, expand=FALSE, scale-factor=2, border-width=0, resize-mode, child, type=GTK_WINDOW_TOPLEVEL, title=NULL, role=NULL, resizable=TRUE, modal=FALSE, window-position=GTK_WIN_POS_NONE, default-width=-1, default-height=-1, destroy-with-parent=FALSE, hide-titlebar-when-maximized=FALSE, icon, icon-name=NULL, screen, type-hint=GDK_WINDOW_TYPE_HINT_NORMAL, skip-taskbar-hint=FALSE, skip-pager-hint=FALSE, urgency-hint=FALSE, accept-focus=TRUE, focus-on-map=TRUE, decorated=TRUE, deletable=TRUE, gravity=GDK_GRAVITY_NORTH_WEST, transient-for, attached-to, has-resize-grip, resize-grip-visible, application, is-active=FALSE, has-toplevel-focus=FALSE, startup-id, mnemonics-visible=FALSE, focus-visible=FALSE, is-maximized=FALSE)

julia> 
(<unknown>:40503): Gtk-WARNING **: Error loading theme icon 'document-open' for stock: Icon 'document-open' not present in theme

[...]

PR #327, opened 2017-09-19
update gdk-pixbuf cache upon installation

tlnagy:
This should fix https://github.com/JuliaImages/ImageView.jl/issues/127 and #180. Not sure what's the best way to test whether this is working successfully since I had already fixed this error locally.


tlnagy:
I was able to get a `Pkg.build("Gtk")` to work successfully, but not sure if my previous changes were wiped when that was run.

tlnagy:
Bump @timholy, @tknopp

Issue #324, opened 2017-08-25
Precompilation fails on Julia master


Cody-G:
Error message:

```julia
julia> using Gtk
INFO: Precompiling module Gtk.
ERROR: LoadError: LoadError: LoadError: syntax: malformed expression
Stacktrace:
 [1] include_relative(::Module, ::String) at ./loading.jl:464
 [2] include at ./sysimg.jl:14 [inlined]
 [3] include(::String) at /home/cody/.julia/v0.7/Gtk/src/GLib/GLib.jl:1
 [4] include_relative(::Module, ::String) at ./loading.jl:464
 [5] include at ./sysimg.jl:14 [inlined]
 [6] include(::String) at /home/cody/.julia/v0.7/Gtk/src/Gtk.jl:3
 [7] include_relative(::Module, ::String) at ./loading.jl:464
 [8] include(::Module, ::String) at ./sysimg.jl:14
 [9] anonymous at ./<missing>:2
while loading /home/cody/.julia/v0.7/Gtk/src/GLib/../../deps/ext_glib.jl, in expression starting on line 9
while loading /home/cody/.julia/v0.7/Gtk/src/GLib/GLib.jl, in expression starting on line 66
while loading /home/cody/.julia/v0.7/Gtk/src/Gtk.jl, in expression starting on line 8
ERROR: Failed to precompile Gtk to /home/cody/.julia/lib/v0.7/Gtk.ji.
Stacktrace:
 [1] compilecache(::String) at ./loading.jl:598
 [2] _require(::Symbol) at ./loading.jl:407
 [3] require(::Symbol) at ./loading.jl:318
```

tknopp:
yes Gtk.jl is not ready working on Julia master

tknopp:
it seems an issue with 
```
    macro assign_if_unassigned(expr)
        # BinDeps often fails and generates corrupt deps.jl files
        # (https://github.com/JuliaLang/BinDeps.jl/issues/146),
        # but most of the time, we don't care
        @assert expr.head === :(=)
        left = expr.args[1]
        right = expr.args[2]
        quote
            if !isdefined(current_module(), $(QuoteNode(left)))
                global const $(esc(left)) = $(esc(right))
            end
        end
    end
```

@vtjnash: Do you have an idea?

fcard:
Looks like a julia bug. Minimal example:
```julia
macro m()
  :(global $(esc(:x)) = 1)
end
```
```julia
julia> @m
[...]

PR #331, opened 2017-10-24
Work towards making package work on nightly

timholy:
This doesn't get Gtk building on 0.7, but at least it's a step in the right direction. For now I need to shift to other things.

As long as folks are fine with bumping the minimum Julia requirement to 0.6, I think this can be merged (if CI passes). We could then run femtocleaner, which might be nice for getting rid of a whole slew of warnings.


tknopp:
yes lets drop 0.5 support. Trying to be compatible with 0.5 and nightly is no fun.
And thanks for working on this!

tknopp:
CI did not pass. `WORD_SIZE` is not defined

timholy:
We should also hold off on this because of https://github.com/JuliaGraphics/Gtk.jl/issues/324#issuecomment-338974728. It may be a couple days before I get back to this, we'll see.

ihnorton:
small addendum

```
diff --git a/src/GLib/gtype.jl b/src/GLib/gtype.jl
index cbdafb6..271caf7 100644
--- a/src/GLib/gtype.jl
+++ b/src/GLib/gtype.jl
@@ -220,7 +220,7 @@ function get_type_decl(name, iname, gtyp, gtype_decl)
         end
         gtype_wrappers[$(QuoteNode(iname))] = $ename
         macro $einame(args...)
-            Base.depwarn(string("\@", $einame, " is deprecated, use ", $einame, " instead."), :Gtk)
+            Base.depwarn(string("@", $einame, " is deprecated, use ", $einame, " instead."), :Gtk)
             Expr(:call, $ename, map(esc, args)...)
         end
         function $einame(args...; kwargs...)
```

lobingera:
... just to mention this:
```
ERROR: LoadError: LoadError: LoadError: MethodError: no method matching g_type(::Expr, ::Expr, ::Expr)
Closest candidates are:
  g_type(::Symbol, ::Any, ::Expr) at /home/lobi/.julia/v0.7/Gtk/src/GLib/gtype.jl:118
  g_type(::Symbol, ::Any, ::Symbol) at /home/lobi/.julia/v0.7/Gtk/src/GLib/gtype.jl:108
Stacktrace:
 [1] @Gtype(::LineNumberNode, ::Module, ::Any, ::Any, ::Any) at /home/lobi/.julia/v0.7/Gtk/src/GLib/gtype.jl:239
 [2] include_relative(::Module, ::String) at ./loading.jl:509
 [3] include at ./sysimg.jl:15 [inlined]
 [4] include(::String) at /home/lobi/.julia/v0.7/Gtk/src/Gtk.jl:3
 [5] top-level scope
 [6] include_relative(::Module, ::String) at ./loading.jl:509
 [7] include(::Module, ::String) at ./sysimg.jl:15
 [8] top-level scope
 [9] top-level scope at ./<missing>:2
in expression starting at /home/lobi/.julia/v0.7/Gtk/src/gtktypes.jl:18
in expression starting at /home/lobi/.julia/v0.7/Gtk/src/gtktypes.jl:36
in expression starting at /home/lobi/.julia/v0.7/Gtk/src/Gtk.jl:52
ERROR: Failed to precompile Gtk to /home/lobi/.julia/lib/v0.7/Gtk.ji.
Stacktrace:
 [1] compilecache(::String) at ./loading.jl:637
 [2] _require(::Symbol) at ./loading.jl:448
 [3] require(::Symbol) at ./loading.jl:300
```

ihnorton:
That's `https://github.com/JuliaLang/julia/issues/23221` (I dipped my toe in and quickly got bogged down, then ran out of time)

lobingera:
@ihnorton Looking at gtype.jl this seems anyway a place where metaprograming is stretched - sometimes by far. Interesting are the comments about "0.2 compatibility" ...

tknopp:
How can we make this package maintainable again? The high level things are no issue but the `gtype` things are too complicated for me...  @vtjnash: Would you have time to make Gtk.jl working again on 0.7? This would be sooo great!

jonathanBieler:
Hurrah !

nice

The tests don't look good but at least I got a Window to show up.

The macro issue isn't too difficult to fix, but after that I got issues with MutableTypes and unsafe_converts, I'm not sure I fixed them the correct way. I'll keep investigating.

https://github.com/JuliaGraphics/Gtk.jl/blob/master/src/GLib/MutableTypes.jl#L32 (this look quite dirty)

tknopp:
great!!! Thanks a lot for working on this. I have still not yet tried 0.7 nightly since without Gtk.jl working this is not useful for me.

tknopp:
@jonathanBieler: Do you have your work on making Gtk work on 0.7 on some branch?

jonathanBieler:
I should have it somewhere yes, I can upload it if you want. I wanted to work on it but I haven't had much time, I also had to fix some upstream issues with Homebrew on OS X.

tknopp:
maybe the easiest would be integrating it directly into the `teh/partial0.7` branch. Do you have access?

jonathanBieler:
I don't have any special access no.

tknopp:
now you have.

jonathanBieler:
Alright, I think my git history is a bit messed up, I'll have to clean things up and then I'll try to push my changes.

jonathanBieler:
The commits are a bit messy because I converted all the Void to Nothing (not sure why the gbox3 didn't merge). It doesn't build online but on my machine there's only 2-3 tests failing.

tknopp:
the gbox3 is since I merged a PR that changed the file. This will need a rebase.

tknopp:
```
julia> @time using Gtk
164.628849 seconds (244.12 M allocations: 14.633 GiB, 3.08% gc time)
```

wow

tknopp:
The CI failures seem to be unrelated. Does this pass test @jonathanBieler? Since 0.6 is still working on the branch we could also merge your current status to master.

jonathanBieler:
Still 3 tests failing, I'll try to fix them this week.

jonathanBieler:
I've looked at the last errors but if I understand the problem correctly we need to take a design decision for the setproperty!/getproperty business (#344) to solve them.

The main issue is that we have:

    setproperty!(w::GObject, name::Symbol, value) 

which in 0.7 is called when doing `w.name = value`

One solution is to not extend `Base.setproperty!` in Gtk, but then we would
need to call `Gtk.setproperty!` outside of Gtk to set Gtk properties and
it kind of defeats the purpose of Base.setproperty!.

But if we extend base then:

    w.handle = value

Would try to access a Gtk property called `handle` instead of the Julia
field, which I think would be very confusing. I've tried a solution
were `Base.setproperty!` access first Julia fields if they exist,
and Gtk property otherwise but I'm getting some nasty bugs with it.

A simple solution would be to have a new method to access Gtk properties
(e.g. 'setgtkproperty!'), but of course it would be nicer to take advantage
of the new syntax opportunities.

The only solution I can think of right now is to have something like:

    setproperty!(w::GObject, name::Symbol, value::GValue)

Where `GValue` is some light wrapper, such that to set a Gtk property you would do:

    w.handle = GValue(value)

And to set a Julia field:

    w.handle = value

tknopp:
I don't see the issue. 
```
The syntax a.b = c calls setproperty!(a, :b, c)
```
so this is actually right isn't it? b is the property name (like `:value`), `a` is the widget and `c` is the value to be set. I do not understand what the handle is doing.

tknopp:
`getindex` is trickier though

tknopp:
ping @vtjnash: @jonathanBieler is making Gtk.jl 0.7 ready and your input to the `getproperty` / `setproperty!` business would be very valuable. (See comment above). I actually don't have an idea how getproperty should be defined now.

jonathanBieler:
The issue is that you can have types that are children of GObject with Julia fields, e.g.

https://github.com/JuliaGraphics/Gtk.jl/blob/master/src/buttons.jl#L58

For these if you do :

GtkRadioButtonGroup.handle = value

You will try to modify and Gtk property called :handle instead of the Julia field.

tknopp:
As far as I have understood it, that is why there is https://docs.julialang.org/en/latest/base/base/#Core.getfield and the corresponding `setfield!`
In case we have overloaded `getproperty`, the `getfield` method gives us a way to access the actual field.

vtjnash:
My goal was always to connect this to the `.` syntax, so the convergence of name was vaguely intentional.

```julia
obj.field[Type] # currently getproperty(obj, :field, Type)
obj.field[Type] = bar # currently setproperty!(obj, :field, Type, bar)
obj.field = bar # currently setproperty!(obj, :field, bar)
```

Basic sketch of implementation of above:
```julia
struct FieldRef{T}
    obj::T
    field::Symbol
    global getproperty(obj::T, field::Symbol) where {T} = new{T}(obj, field)
end
getindex(::FieldRef, ::Type{T}) where {T} = ccall
setindex!(::FieldRef, val, ::Type{T}) where {T} = ccall
setproperty!(obj, field::Symbol) = ccall
```

tknopp:
@vtjnash and what about those issues with composite types claiming to be GObjects? Those would require `Core.getfield` and `Core.setfield!`, right? (see above the issue reported by Jonathan)

tknopp:
yes true. At least until we have full introspection and know the type of the field. That will be the point where this interface will fight with GAccessor. Both provide basically the same functionality.

jonathanBieler:
I'm looking into that again but if I understand properly the design above, `obj.field` would return a `FieldRef` instead of the Julia field, which would break code like this https://github.com/JuliaGraphics/Gtk.jl/blob/86c391eed8bf0591ee636c2f77e6465be1b9b995/src/cairo.jl#L40

vtjnash:
We could call `fieldname !== :handle && isdefined(x, fieldname)` in the getproperty definition to make this convenient

jonathanBieler:
So give priority to Julia fields ? One issue is that in a function like this:

```julia
function foo(x::GObject)
    val = x.bar #let's hope this is a Gtk property
    # do something with val ...
end 
```

Will suddenly break if I add a "bar" field to my object (since `x.bar` would now refer to the Julia field instead of the Gtk one).

In my opinion it should be be clear at the syntax level whether you want to set/get a Julia field or a Gtk property.

vtjnash:
I expect that subtyping GObject should be rare, but that overriding the Gtk property field in that way is valid, when that's what you want.

jonathanBieler:
I guess it could work, at least it wouldn't break existing code since currently `x.bar` always refer to a Julia field.

test/misc.jl
```diff
@@ -36,9 +36,9 @@ are expected and a sign of normal operation.
 c = GtkCanvas()
 win = GtkWindow(c)
 showall(win)
-@guarded draw(c) do widget
-    error("oops")
-end
+# @guarded draw(c) do widget
```
jonathanBieler: This needs to be fixed.

test/gui.jl
```diff
@@ -602,25 +617,26 @@ push!(tv,c1)
...
-@test isvalid(ls, iter) == false
-
-tmSorted=TreeModelSort(ls)
-G_.model(tv,tmSorted)
-G_.sort_column_id(TreeSortable(tmSorted),0,GtkSortType.ASCENDING)
-it = convert_child_iter_to_iter(tmSorted,Gtk.iter_from_index(ls, 1))
-select!(selmodel, it)
-iter = selected(selmodel)
-@test TreeModel(tmSorted)[iter, 1] == 35
+if false #this crashes
```
jonathanBieler: This needs to be fixed.

test/gui.jl
```diff
@@ -637,11 +653,12 @@ c1=TreeViewColumn("A", r1, Dict([("text",0)]))
 push!(tv,c1)
 w = Window(tv, "Tree View")|>showall
 
-
+if false #this crashes
```
jonathanBieler: This needs to be fixed.

jonathanBieler:
This last commit implements this new syntax, the first thing I did was to rename `getproperty` and `setproperty!` to `get_gtk_property` and `set_gtk_property!`, which are meant to be internal functions to do the ccall that gets/sets the Gtk properties. It also allows for an easy transition (it's just a search/replace).

I haven't made use of the new syntax except in a few tests, but it seems to work.

I've disabled a few tests that broke and need investigating:

```
 select!(selmodel, Gtk.iter_from_index(ls, 1))
 (<unknown>:93228): Gtk-CRITICAL **: 17:16:00.879: gtk_list_store_get_path: assertion 'iter->stamp == priv->stamp' failed
```

At this point it would be nice if it was possible to run FemtoCleaner, because the amount of noise and slowness from deprecations is quite annoying.  I don't know if it's possible though.

tknopp:
I have installed femtocleaner. Have never used it though. Maybe we have to merge your branch first?

tknopp:
and using the internal getter/setter seems like a good solution. If everything is passing we should migrate the examples to the new syntax

src/GLib/MutableTypes.jl
```diff
@@ -29,7 +29,7 @@ mutable{T <: Ptr, N}(x::Array{T, N}, i = 1) = mutable(x[i])
 mutable{T}(x::Ptr{T}, i = 1) = x + (i - 1) * sizeof(T)
 mutable{T}(x::T, i) = (i == 1 ? mutable(x) : error("Object only has one element"))
 
-_addrof{T}(b::T) = ccall(:jl_value_ptr, Ptr{T}, (Ptr{Any},), &b)
+_addrof{T}(b::T) = pointer_from_objref(b)
```
jonathanBieler: @vtjnash is this correct ?

src/GLib/gvalues.jl
```diff
@@ -161,17 +162,32 @@ function getproperty{T}(w::GObject, name::AbstractStringLike, ::Type{T})
...
+        new{T}(obj, field)
+    end
+end
+
+getindex(f::FieldRef, ::Type{T}) where {T} = get_gtk_property(f.obj,f.field,T) 
+
+function setindex!(f::FieldRef, value::T, ::Type{T}) where {T}
+    isdefined(f.obj,f.field) && return setfield!(f.obj, f.field, value)
+    set_gtk_property!(f.obj, f.field, value) 
+end
```
vtjnash: `return f`

src/GLib/MutableTypes.jl
jonathanBieler: The iters error seems to be caused by invalid GtkTreeIter when doing this:

```
ls=ListStore(Int32,Bool)
push!(ls,(44,true))
push!(ls,(33,true))
insert!(ls, 2, (35, false))
```

insert! gets a valid iter from here, but it seems to become invalid at the insert time. I'm not sure what's going on.

https://github.com/JuliaGraphics/Gtk.jl/blob/master/src/lists.jl#L106

vtjnash: Yes

src/GLib/MutableTypes.jl
```diff
@@ -29,7 +29,7 @@ mutable{T <: Ptr, N}(x::Array{T, N}, i = 1) = mutable(x[i])
...
 Base.cconvert{P <: Ptr, T}(::Type{P}, b::MutableX{T}) = isbits(T) ? convert(P, _addrof(b)) : convert(P, _addrof(b.x))
```
vtjnash: `isbitstype`

Should also trigger a femtobot request on the repo, to update all of the `f{T}()` syntax to `f() where {T}`

tknopp: I had installed femtobot but it does not seem to be triggered. Is that actually possible on a branch? Or should this first be merged in order to be able running femtobot?

jonathanBieler:
I think it's better to do a new branch once this is merged otherwise the diffs here will be unreadable.

I think I solved that last issue, the ccall signature for `GtkTreeIter` wasn't working properly (`Ptr{GtkTreeIter}` instead of `Ref{GtkTreeIter}`). In order to be able to pass both `Mutable{GtkTreeIter}` and `GtkTreeIter` to these functions I added two methods to `cconvert` to automatically convert to the right type. 

All tests run on my computer, but I might have missed some functions that are not tested or made some mistakes. GtkTextIter's will probably have the same issue.

tknopp:
@jonathanBieler: We could merge and proceed from there. Then we could run femtocleaner which seems to just run over master.

vtjnash:
Femtocleaner ran now, so you can accept its fixes on the PR. Good find on the ccall signature. Eventually might be helpful to switch to using Ref (this type and interface inspired and influenced its creation)

jonathanBieler:
I'm getting a segfault on macOS, so there's still some issues. I'll try to narrow it down.

tknopp:
I made some further progress in getting this package 0.7 ready but there is a very basic issue that needs attention:

execute:
```
using Gtk

win = GtkWindow("My First Gtk.jl Program", 400, 200)

b = GtkButton("Click Me")
push!(win,b)

function on_button_clicked(w)
  println("The button has been clicked")
end
signal_connect(on_button_clicked, b, "clicked")

showall(win)
```
and click on the button. Then I get:
```
ERROR: StackOverflowError:
Stacktrace:
 [1] convert(::Type{GObject}, ::Ptr{GObject}) at /Users/knopp/.julia/dev/Gtk/src/GLib/gtype.jl:299 (repeats 80000 times)
```
So there is a problem with callbacks. ping @vtjnash if you find time.

jonathanBieler:
That seems like a circular definition, I will have some time in one week, I'll look into that if it hasn't been solved.

tknopp:
tests are passing now! will merge

PR #341, opened 2017-12-13
switch `build.jl` to use libgtk-3-dev from apt

### LINKED ISSUE #289: build issues on Ubuntu 16.04 (wrong apt package specified?)
I found that, on Ubuntu 16.04, the package only builds correctly when `libgtk-3-dev` is installed from apt-get, while the deps script tries to install `libgtk-3-0`.

Looking at their contents, `libgtk-3-0` provides the file `/usr/lib/x86_64-linux-gnu/libgtk-3.so.0`, while `libgtk-3-dev` provides `/usr/lib/x86_64-linux-gnu/libgtk-3.so`. Only the latter makes the build script happy. (In fact, they are both just symlinks to the same file though.)

Furthermore, when not finding the desired library the script tried to invoke `sudo apt-get install libgtk-3-0` (which would not have worked anyway for the reasons stated above, in fact it was installed already on my system), which resulted in the following error:

```
INFO: Building Gtk
Installing dependency libgtk-3-0 via `sudo apt-get install libgtk-3-0`:
sudo: no tty present and no askpass program specified
======================================================================[ ERROR: Gtk ]=======================================================================

LoadError: failed process: Process(`sudo apt-get install libgtk-3-0`, ProcessExited(1)) [1]
while loading /home/carlo/.julia/v0.6/Gtk/deps/build.jl, in expression starting on line 61
```

I guess this may be a problem with BinDeps, though?

rdeits:
Resolves #289 

To verify that this works, I created a new Dockerfile based on [docker-library/julia](https://github.com/docker-library/julia/blob/06148bd09222645c2996f50094d76aeeb9ed4556/Dockerfile ) which you can find here: https://gist.github.com/rdeits/b92935c572f9d057d7121c904e2c8040 The dockerfile has no gtk packages pre-installed, but builds Gtk.jl successfully after this change. 

I've tested the Docker build on Ubuntu 14.04, 16.04, and 17.10 along with Debian Jessie.

If you'd like, I can set up the Docker build as an additional build on Travis so this will get tested in the future.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -17,7 +17,7 @@ deps = [
 ]
 
 if @compat is_linux()
-    provides(AptGet, "libgtk-3-0", deps)
+    provides(AptGet, "libgtk-3-dev", deps)
     provides(Yum, "gtk3", deps)
 end
 


tknopp:
thanks

PR #342, opened 2017-12-13
WIP: Add docker build

rdeits:
Demonstrates adding Docker build from #341 

Probably won't work right away, since I'm guessing about the Travis commands a little bit.

### DIFF
--- a/travis-script.sh
+++ b/travis-script.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+set -ex
+
+if [[ -a .git/shallow ]]; then 
+	git fetch --unshallow
+fi
+if [[ "$DOCKER_BUILD" = true ]]; then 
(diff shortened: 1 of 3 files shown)


travis-script.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+set -ex
+
+if [[ -a .git/shallow ]]; then 
+	git fetch --unshallow
```
vtjnash: I'm not sure why we need git history, but perhaps it would be faster to set this from the travis file:
```
git:
  depth: 2147483647
```

rdeits: Yeah, that's probably left over from the old PkgDev templates which did this. Your way is nicer.

rdeits:
Ok, docker build is working now.

PR #343, opened 2017-12-15
fixes interger parsing on index_from_iter

### LINKED ISSUE #337: Unable to use index_from_iter, because it parses string to integer in the wrong (old) way
Using index_from_iter throws:
ERROR: MethodError: Cannot `convert` an object of type String to an object of type Int64

https://github.com/JuliaGraphics/Gtk.jl/blob/4e80acb8256b33858b6db90e7aa9fa76059ab32b/src/lists.jl#L127

Now Julia imposes the parse function, as:
`index_from_iter(store::GtkListStore, iter::TRI) = parse(Int, get_string_from_iter(GtkTreeModel(store), iter)) + 1`

tibaes:
fixes #337


tknopp:
nice! Could you also add a test so that this function will always work correctly?

Here:

https://github.com/JuliaGraphics/Gtk.jl/blob/master/test/gui.jl#L591

I had introduced tests for the list store. Just add a row that tests your function.

PR #354, opened 2018-06-02
Drop support for Julia 0.5

Vexatos:
Julia 0.5 has been declared officially unmaintained, so [package updates won't be accepted if they support this version](https://github.com/JuliaLang/METADATA.jl/pull/12729). This led to JuliaLang/METADATA.jl#14843 getting rejected. This PR makes 0.6 the minimum required julia version and removes compat macros that aren't needed anymore.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ os:
   - linux
   - osx
 julia:
-  - 0.5
   - 0.6
   - nightly
 notifications:
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,7 +1,7 @@
-julia 0.5
+julia 0.6
 Cairo 0.2.2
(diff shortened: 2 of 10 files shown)


tknopp:
Thanks

tknopp:
I really wonder what one should do now. Do a new release?

Vexatos:
Well the old release did not get merged into METADATA.jl, so it essentially did nothing.

tknopp:
yes but is it correct that I need to do a new release (such that your change is included)?

Vexatos:
Most likely, yes, since the tags are meant to be static.

tknopp:
ok, I have made a new release: https://github.com/JuliaLang/METADATA.jl/pull/14965

PR #366, opened 2018-08-23
Update FileFilter to GtkFileFilter and correct filter spacing in Docs

jordancluts:
Updates FileDialog documentation so that the examples work as expected.


tknopp:
thanks. I have seen that there are other places where this is not consistent:

http://juliagraphics.github.io/Gtk.jl/latest/manual/signals.html

In the long run, I would like to make the ShortNames module the default and remove all the Gtk prefixes.

PR #373, opened 2018-09-02
Update section 'Builder and Glade' of documentation

Element-126:
Only minor fixes here:
* Add missing quotes to the file path.
* The note on `Pkg.dir("MyPackage")` should maybe also be modified to reflect the fact that it is deprecated in 1.0. I am not sure which replacement is best here, though. I tend to use `@__DIR__` a lot, but the deprecation warning recommends using `joinpath(dirname(pathof(MyPackage)), "..", paths...)` instead. I have *not* changed the note.

### DIFF
--- a/docs/src/manual/builder.md
+++ b/docs/src/manual/builder.md
@@ -52,7 +52,7 @@ the extension `.glade`. Lets assume we have created a file `myapp.glade` that lo
 In order to access the widgets from Julia we first create a `GtkBuilder` object that will serve as our
 connector between the XML definition and our Julia code.
 ```julia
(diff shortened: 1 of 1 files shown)


tknopp:
`@__DIR__` should be used in that context. I usually put my glade files into a `builder` directory within `src/` and then one can use
```
  uifile = joinpath(@__DIR__,"builder","myuifile.ui")
```

Element-126:
@tknopp Could you add that to the note ?

tknopp:
I don't have time to work on the docs right now. So help is appreciated

Element-126:
I will do it then.

PR #376, opened 2018-09-02
Update section 'List and Tree Widgets'

Element-126:
* Add missing `Gtk`'s
* Remove unnecessary `enumerate` on line 83.
* Fix typos.
* Fix broken example in lines 101:103. For some reason I need to use 1-based indices to make all three columns reorderable. Not sure why this is the case here and not elsewhere since I am not very familiar with GTK.
* `contains(a, b)` -> `occursin(b, a)` in Julia 1.0.

### DIFF
--- a/docs/src/manual/listtreeview.md
+++ b/docs/src/manual/listtreeview.md
@@ -75,13 +75,13 @@ push!(tv, c1, c2, c3)
 ```
 Then we can display the tree view widget in a window
 ```julia
-win = Window(tv, "List View")
+win = GtkWindow(tv, "List View")
 showall(win)
 ```
 If you prefer that the columns are resizable by the user call
 ```julia
-for (i,c) in enumerate([c1,c2,c3])
-  GAccessor.resizable(c,true)
+for c in [c1, c2, c3]
+    GAccessor.resizable(c, true)
 end
 ```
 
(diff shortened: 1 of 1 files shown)


tknopp:
We do not support 0.6 anymore. Therefore you can remove the `contains` entirely. I am still merging.

Kelvyn88:
@JLTastet there is a bug. If I type a letter not contained in the words available in "ls", then the window suddenly closes and julia session ends.

JLTastet:
> @JLTastet there is a bug. If I type a letter not contained in the words available in "ls", then the window suddenly closes and julia session ends.

It seems that `Gtk.jl` changed quite a lot since this PR. I cannot reproduce your problem when using the same version I tested the documentation on (`v0.16.2`). However, with the most recent versions of Gtk and its dependencies, the example that you mention indeed caused Julia to freeze (but not crash) _once_, but the bug did not repeat.

In case it matters, I get the following warnings when including `Gtk`:
```
Gtk-Message: 21:41:51.083: Failed to load module "canberra-gtk-module"
Gtk-Message: 21:41:51.083: Failed to load module "pk-gtk-module"
Gtk-Message: 21:41:51.083: Failed to load module "canberra-gtk-module"
Gtk-Message: 21:41:51.083: Failed to load module "pk-gtk-module"
```

I am not familiar with the latest Gtk, so I doubt I can help here.

PR #379, opened 2018-09-04
Fix examples

wookay:
I tried to fix the examples to run in Julia 0.7, 1.0.
Some remaining annoying things.

* 2048.jl
  -  Got the following message on`info_dialog`
 ```Gtk-Message: 13:40:21.376: GtkDialog mapped without a transient parent. This is discouraged.```

* gl-area.jl
  - Displayed the text `Not implemented on OS X` on GLArea in my mac OSX. I can't find where it has to be.


tknopp:
Thanks,
The first message is ok since you don't have a top-level window.
Unfortunately GLArea is not implemented on OS X. So nothing we can do there.

PR #383, opened 2018-09-04
Fix build.jl on Windows

Vexatos:
`Pkg` now has to be explicitly imported to be used.


tknopp:
I will fix the Pkg.dir warning in a next step, thanks!

PR #385, opened 2018-09-13
Fix GtkTextIter and add tests

jonathanBieler:
There was a couple of issues with those, similar to the ones with `GtkTreeIter`. I also added a bunch of tests.


tknopp:
nice, thanks!

PR #386, opened 2018-09-13
Change `curr_module` to `__module__` (or `@__MODULE__` when appropriate)

### LINKED ISSUE #367: Defining gtktype outside of Gtk doesn't work anymore
```julia
julia> import ..Gtk: suffix, @gtktype_custom_symname,
       @gtktype_custom_symname_and_lib, @Gtype, _gtksubtype_constructors

julia> Gtk.@gtktype GtkClipboard
ERROR: UndefVarError: GtkClipboardLeaf not defined
Stacktrace:
 [1] _gtksubtype_constructors(::Symbol) at /Users/bieler/.julia/dev/Gtk/src/gtktypes.jl:4
 [2] top-level scope at /Users/bieler/.julia/dev/Gtk/src/GLib/gtype.jl:199
 [3] top-level scope at /Users/bieler/.julia/dev/Gtk/src/gtktypes.jl:18
```

I think the issue is with the new behaviour of `@__MODULE__` here, it tries to get GtkClipboardLeaf from Gtk instead of the calling module.

https://github.com/JuliaGraphics/Gtk.jl/blob/66ca486d6c15199ff027afa833653c4bf217ea24/src/gtktypes.jl#L2-L4

The solution might be to pass the calling module as an optional argument, I'll see if I can do something along these lines.

fcard:
Fixes #367. Sorta messy but I didn't want to change things around too much. Still needs a test. Maybe `cm` needs a better name?

### DIFF
--- a/src/GLib/GLib.jl
+++ b/src/GLib/GLib.jl
@@ -18,17 +18,14 @@ import Base: convert, copy, show, size, length, getindex, setindex!, get,
 using Libdl
 
 export GInterface, GType, GObject, GBoxed, @Gtype, @Gabstract, @Giface
-export GEnum, GError, GValue, gvalue, make_gvalue, g_type
+export GEnum, GError, GValue, gvalue, make_gvalue, @make_gvalue, g_type
 export GList, glist_iter, _GSList, _GList, gobject_ref, gobject_move_ref
 export signal_connect, signal_emit, signal_handler_disconnect
 export signal_handler_block, signal_handler_unblock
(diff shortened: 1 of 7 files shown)


jonathanBieler:
Thanks, I'll test it tomorrow.

tknopp:
@fcard: thanks a lot, this is much better than my `curr_module` hack and prevents the warning. I will merge this as is (since tests are passing) and maybe @jonathanBieler can test if this now is also working within an external module.

PR #395, opened 2018-10-25
g_timeout_add and other things

jonathanBieler:
I've moved `g_timeout_add`, `g_idle_add`, and a couple of other things that I had in GtkExtensions.jl. I still have other things in there (text and tree related mainly) but I need to review them.

I've added tests for most of it, but no documentation.

### DIFF
--- a/src/GLib/GLib.jl
+++ b/src/GLib/GLib.jl
@@ -22,6 +22,7 @@ export GEnum, GError, GValue, gvalue, make_gvalue, @make_gvalue, g_type
 export GList, glist_iter, _GSList, _GList, gobject_ref, gobject_move_ref
 export signal_connect, signal_emit, signal_handler_disconnect
 export signal_handler_block, signal_handler_unblock
+export g_timeout_add, g_idle_add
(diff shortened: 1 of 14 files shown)


tknopp:
there are different test errors. The one on windows might be because the sleep needs to be longer?

jonathanBieler:
There's probably an issue with this, I copied it from the signal_connect part but it might not be correct. I can reproduce the issue on Windows so I'll try to fix it.

https://github.com/JuliaGraphics/Gtk.jl/pull/395/files#diff-8c2b506e7179e52e816713686d731068R374

jonathanBieler:
I'm not 100% sure about those `g_timeout_add`, but it seems to work for the most part.

stevengj:
Now that Julia is getting full multi-threaded support in 1.3, you probably should add the thread-safe API function `gdk_threads_add_idle_full`.

tknopp:
I think the `gdk` versions are deprecated, see
https://gitlab.gnome.org/GNOME/gtk/commit/334acbfc39cc08869932ec046e4d13e6c4b64be6

PR #402, opened 2019-01-13
Added Alpha to RGBA struct constructor

Cruor:
Fixes bug with (seemingly) random alpha channel on alpha enabled Pixbufs.
Makes convert(::RGBA, ::Unsigned) work for Pixbuf data manipulation.


tknopp:
looks good, thanks

Issue #400, opened 2019-01-09
unable to use `path` in list.jl (calling `gtk_tree_model_get_path`)


zweiglimmergneis:
Calling `path` in list.jl I get

> (julia103:6466): Gtk-CRITICAL **: 12:24:06.326: gtk_list_store_get_path: assertion 'iter->stamp == priv->stamp' failed
> (julia103:6466): Gtk-CRITICAL **: 12:24:06.326: gtk_tree_path_copy: assertion 'path != NULL' failed
ERROR: LoadError: objects of type getfield(Gtk, Symbol("##257#258")) cannot be finalized

Example:
``` julia
window = GtkWindow("issue path", 200, 100)
boxtop = GtkBox(:v) # vertical box, basic structure

wscroll = GtkScrolledWindow()

# create ls::GtkListStore and tv::GtkTreeView
ls, tv = itemlist([Int, String], ["No", "Name"])

push!(wscroll, tv)
push!(boxtop, wscroll)
push!(window, boxtop)

showall(window);

push!(ls, (3, "London"))
iter = push!(ls, (7, "Grenoble"))

pathend = Gtk.path(GtkTreeModel(ls), iter)
```
My overall goal is to make the last row in `ls` visible with a call to `gtk_tree_view_scroll_to_cell`, which works in C.

tknopp:
thanks for the report. It seem the entire path handling is not tested. We thus need tests and fixes (to your problem.

I have found:
https://github.com/tknopp/Julietta.jl/blob/a90bb40042089157e490438f640404a3d256e6f1/src/workspace.jl#L42
so that at least something was working at a certain time.

A pull request would be very much appreciated.

jonathanBieler:
I also got some issues with Trees today, there was some deprecation that weren't fixed. I still have some warning with GtkTreeIters and we should add some tests indeed. I can do it when I have time (next week probably).

tknopp:
some tests would be very good!

tknopp:
this was fixed by #401

PR #403, opened 2019-01-25
WIP. Adding tests for GtkTrees

jonathanBieler:
Following #400 I'm adding some tests for GtkTreeViews. 

I fixed a couple of issues but I'm getting segfaults on `selected_rows`, calling `gtk_tree_selection_get_selected_rows` seems to work but when I wrap the resulting pointer into ` GLib.GList` and try to do something with it (even show it) it crashes. We need to investigate further before we can merge this.

### DIFF
--- a/src/GLib/glist.jl
+++ b/src/GLib/glist.jl
@@ -65,8 +65,9 @@ end
 done_(::LList, s::LListPair{L}) where {L <: _LList} = done_(s[1], s[2])
 
 ## Standard Array-like declarations
+show(io::IO, ::MIME"text/plain", list::GList{L, T}) where {L, T} = show(io, list)
 show(io::IO, list::GList{L, T}) where {L, T} = print(io, "GList{$L => $T}(length = $(length(list)), transfer_full = $(list.transfer_full))")
-# show{L, T}(io::IO, list::Type{GList{L, T}}) = print(io, "GList{$L => $T}")
+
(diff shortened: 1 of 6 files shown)


jonathanBieler:
I've been looking at this but I don't understand what's going on. It looks like the first thing that is crashing is trying to access an item of the list with `deref` :

```julia
ptr = ccall((:gtk_tree_selection_get_selected_rows, Gtk.libgtk),
Ptr{Gtk._GList{Gtk.GtkTreePath}},
(Ptr{GObject}, Ptr{Ptr{GtkTreeModel}}),
selection, C_NULL)

function test(ptr) 
    list = Gtk.GLib.GList(ptr)
    item = Gtk.GLib.nth_last(list)
    #x = Gtk.GLib.deref(item) #this segfaults
    #x
end
test(ptr) 
```
(copy paste this at the end of `tree.jl` in test)

@vtjnash Is [this](https://github.com/JuliaGraphics/Gtk.jl/blob/master/src/GLib/glist.jl#L43) ? I don't really understand the difference between `unsafe_pointer_to_objref` and `unsafe_load`. Any other idea what the problem could be ?

Alternatively we could just disable `selected_rows` for now, looks like nobody is using it.

jonathanBieler:
I'll merge this, I didn't manage to fix the problem with `selected_rows` but having some tests is still better than not. I also fixed a small bug.

tknopp:
ci error is because of homebrew issues in cairo.

Issue #405, opened 2019-02-07
TreeView "row-activated" callback not working


igor-krawczuk:
Hi, I've been trying to use TreeView + the "row-activated" signal in combination with the toggle renderer to create an "selected" toggle, but my callback doesn't fire. trying a minimal example baed on mostly the code from the docs:

```
using Gtk
ls = GtkListStore(String, Int, Bool)

push!(ls,("Peter",20,false))
push!(ls,("Paul",30,false))
push!(ls,("Mary",25,true))

tv = GtkTreeView(GtkTreeModel(ls))


rTxt = GtkCellRendererText()
rTog = GtkCellRendererToggle()

c1 = GtkTreeViewColumn("Name", rTxt, Dict([("text",0)]))
c2 = GtkTreeViewColumn("Age", rTxt, Dict([("text",1)]))
c3 = GtkTreeViewColumn("Female", rTog, Dict([("active",2)]))

push!(tv, c1, c2, c3)
win = GtkWindow(tv, "List View")
showall(win)

selection = GAccessor.selection(tv)

signal_connect(tv, "row-activated") do widget 
  if hasselection(selection)
     print("logme")
    currentIt = selected(selection)

    # now you can to something with the selected item
    ls[currentIt,3]=!ls[currentIt,3]
  end
end

```

But neither does the toglge change, not does the console log "logme"), whereas if I try e.g. the selection callback from the docs, it does log. I have also tried  `do widget, path, col, udata` [following the signal docs](https://developer.gnome.org/gtk3/stable/GtkTreeView.html#GtkTreeView-row-activated) but it doesn't seem to work that way either (don't get any errors either though).

This is on Linux,  Julia Version 1.0.1, Gtk.jl 0.16.4

jonathanBieler:
I'm not sure what the issue is but using the explicit `signal_connect` solves the problem:

```julia

using Gtk
ls = GtkListStore(String, Int, Bool)

push!(ls,("Peter",20,false))
push!(ls,("Paul",30,false))
push!(ls,("Mary",25,true))

tv = GtkTreeView(GtkTreeModel(ls))

rTxt = GtkCellRendererText()
rTog = GtkCellRendererToggle()

c1 = GtkTreeViewColumn("Name", rTxt, Dict([("text",0)]))
c2 = GtkTreeViewColumn("Age", rTxt, Dict([("text",1)]))
c3 = GtkTreeViewColumn("Female", rTog, Dict([("active",2)]))
[...]

PR #406, opened 2019-02-08
Added cross reference to explicit signal connect to make it easier for newbies to use 'row-activated'

igor-krawczuk:
Since I had the issue myself in #405 and there doesn't seem to be any link between the rendered documentation and the /doc folder  figured this might help future users. (might be an idea to add the (doc folder to the manual in any case? or at least as a "further reading link?)

### DIFF
--- a/doc/more_signals.md
+++ b/doc/more_signals.md
@@ -4,7 +4,7 @@ In addition to the ["simple"
 interface](../README.md#callbacks-and-signals), `signal_connect`
 supports an approach that allows your callback function to be directly
 compiled to machine code.  Not only is this more efficient, but it can
-occasionally be useful in avoiding problems (see issue #161).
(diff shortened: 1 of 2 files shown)


tknopp:
the doc things are obsolete and should be just integrated into docs

igor-krawczuk:
D'accord, is there a place where I can find out which parts are fully obsolete and which ones are still useful (e.g. the explicit signal_connect?) Then I'd try my hand and merging in the useful bits and delete the rest

Le 8 février 2019 17:33:16 GMT+01:00, Tobias Knopp <notifications@github.com> a écrit :
>the doc things are obsolete and should be just integrated into docs

-- 
Envoyé de mon appareil Android avec Courriel K-9 Mail. Veuillez excuser ma brièveté.

tknopp:
i would just start with the one you wanted to link to. Just make a new page and integrate that.

PR #409, opened 2019-03-24
Make length() return Int instead of Int32.

ma-laforge:
`pop!(::GtkListStoreLeaf)` uses `length(::GtkListStore)` which currently returns an `Int32` instead of an `Int`.

It then calls `deleteat!(...)` which expects an index of type `Int` (thus failing to match signature when called with `Int32`).

I am pretty certain that we don't want to "loosen" `deleteat!(...)` to take in `Integer` types... so I instead ensure the output of `length()` is of type `Int`.


tknopp:
looks good to me, thanks

PR #431, opened 2019-08-02
add improved versions of idle_add and timeout_add 

tknopp:
Implementation suggested in https://github.com/JuliaGraphics/Gtk.jl/commit/17bfda1ea78a511cd39d58de56c09cadd93262fb#r34550618

@stevengj could you have a short look? I am by no means a cfunction expert.

@jonathanBieler: Do we want to keep the version with `user_data`?

### DIFF
--- a/src/GLib/signals.jl
+++ b/src/GLib/signals.jl
@@ -360,43 +360,26 @@ function __init__gmainloop__()
     nothing
 end
 
-function g_timeout_add(interval::Integer, cb::Function, user_data::CT) where CT
-
-    callback = @cfunction($cb, Cint, (Ref{CT},) )
-    ref, deref = gc_ref_closure(user_data)
-    
-    return ccall((:g_timeout_add_full, libglib), Cint,
-        (Cint, UInt32, Ptr{Nothing}, Ptr{Nothing}, Ptr{Nothing}),
-        0, UInt32(interval), callback, ref, deref)
-end
-
-_g_timeout_add_callback(cb::Function) = Cint(cb())
-function g_timeout_add(interval::Integer, cb::Function)
-    callback = @cfunction(_g_timeout_add_callback, Cint, (Ref{Function},))
+_g_callback(cb::Function) = Cint(cb())
+function g_timeout_add(cb::Function, interval::Integer)
+    callback = @cfunction(_g_callback, Cint, (Ref{Function},))
     ref, deref = gc_ref_closure(cb)
     return ccall((:g_timeout_add_full, libglib),Cint,
         (Cint, UInt32, Ptr{Cvoid}, Ptr{Cvoid}, Ptr{Cvoid}),
         0, UInt32(interval), callback, ref, deref)
 end
 
-function g_idle_add(cb::Function, user_data::CT) where CT
+@deprecate g_timeout_add(interval, cb, user_data) g_timeout_add(() -> cb(user_data), interval)
 
-    callback = @cfunction($cb, Cint, (Ref{CT},) )
-    ref, deref = gc_ref_closure(user_data)
-
-    return ccall((:g_idle_add_full, libglib),Cint,
-        (Cint, Ptr{Nothing}, Ptr{Nothing}, Ptr{Nothing}),
-        0, callback, ref, deref)
-end
-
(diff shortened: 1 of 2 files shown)


src/GLib/signals.jl
```diff
@@ -370,16 +370,34 @@ function g_timeout_add(interval::Integer, cb::Function, user_data::CT) where CT
...
+_g_timeout_add_callback(cb::Function) = Cint(cb())
+function g_timeout_add(interval::Integer, cb::Function)
+    callback = @cfunction(_g_timeout_add_callback, Cint, (Ref{Function},))
+    ref, deref = gc_ref_closure(cb)
+    return ccall((:g_timeout_add_full, libglib),Cint,
+        (Cint, UInt32, Ptr{Cvoid}, Ptr{Cvoid}, Ptr{Cvoid}),
+        0, UInt32(interval), callback, ref, deref)
+end
+
 function g_idle_add(cb::Function, user_data::CT) where CT
```
stevengj: You could now just implement this as
```jl
g_idle_add(cb::Function, user_data) = g_idle_add(() -> cb(Ref(user_data)))
```
and similarly for `g_timeout_add(interval::Integer, cb::Function, user_data)`.

src/GLib/signals.jl
```diff
@@ -370,16 +370,34 @@ function g_timeout_add(interval::Integer, cb::Function, user_data::CT) where CT
         0, UInt32(interval), callback, ref, deref)
 end
 
+_g_timeout_add_callback(cb::Function) = Cint(cb())
+function g_timeout_add(interval::Integer, cb::Function)
```
stevengj: You might consider making `cb` the first argument so that callers can use `do` syntax.

src/GLib/signals.jl
```diff
@@ -370,16 +370,34 @@ function g_timeout_add(interval::Integer, cb::Function, user_data::CT) where CT
...
     callback = @cfunction($cb, Cint, (Ref{CT},) )
     ref, deref = gc_ref_closure(user_data)
 
-    return ccall((:g_idle_add_full , libglib),Cint,
+    return ccall((:g_idle_add_full, libglib),Cint,
         (Cint, Ptr{Nothing}, Ptr{Nothing}, Ptr{Nothing}),
         0, callback, ref, deref)
 end
 
+_g_idle_add_callback(cb::Function) = Cint(cb())
```
stevengj: You might as well just define a single `_g_callback(cb::Function)` method and use it for both `idle_add` and `timeout_add`, since they are identical.

src/GLib/signals.jl
tknopp: good point, thanks!

stevengj:
You could `@deprecate` the `user_data` version, since it's kind of useless to simulate closures in a language that already has closures.

tknopp:
I changed the code according to your suggestions, thanks. By deprecating the `user_data` version we can also change elegantly the order of arguments in order to move the callback to the front.

jonathanBieler:
I'm fine with removing the `user_data`. My only worry is that the simple and do version of `signal_connect` leads to crashes in some situations, and the robust version of `signal_connect` uses `user_data` in the same way as here.  I don't know if there's a good reason for that or if it's just historical baggage.

https://github.com/JuliaGraphics/Gtk.jl/blob/master/src/GLib/signals.jl#L9

stevengj:
@jonathanBieler, your `signal_connect_generic` problem looks GC-unsafe to me, for the same reason as I commented here (https://github.com/JuliaGraphics/Gtk.jl/commit/17bfda1ea78a511cd39d58de56c09cadd93262fb#r34550746).

You really should change all of the your callbacks to use the static form of `@cfunction` and to pass the function to be called as the "user_data" in the same way as was done here.

jonathanBieler:
@stevengj Just to be clear this was originally written by @vtjnash. I don't understand much of it but I assume there was some reason why it was written so. But if I understand you correctly we should try to refactor it in the same way as here (now we have an inconsistent API for passing user data).

stevengj:
Yes.  User data in a language like Julia should essentially always be passed via closures.   The `user_data` field in the GDK API is a way to simulate closures in C, which is convenient here because it allows us to pass the Julia closure to a static callback function.

PR #438, opened 2019-08-22
Remove OSX from travis

remi-garcia:
For now travis fails because of osx.
```Julia
Building Cairo ───→ `~/.julia/packages/Cairo/p68X8/deps/build.log`
┌ Error: Error building `Cairo`, showing the last 5000 of log: 
...
The command "$TESTCMD -e 'using Pkg; Pkg.build(); Pkg.test(coverage=true)'" exited with 1.
```

To have docs update I understood that we need travis to pass. It might be ok to remove osx temporarily.

I've made a few updates too to avoid a warning:
```Julia
julia> cd(Pkg.dir("Gtk"))
┌ Warning: `Pkg.dir(pkgname, paths...)` is deprecated; instead, do `import Gtk; joinpath(dirname(pathof(Gtk)), "..", paths...)`.
└ @ Pkg.API /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.1/Pkg/src/API.jl:400
```

PR #441, opened 2019-09-06
Put upper bounds on Cairo version

timholy:
This "future-proofs" for new releases of Gtk. Correcting the bounds on prior releases requires a PR directly to the General registry (so this does not suffice on its own).

I also took the opportunity to get rid of Compat.jl, is that OK?


tknopp:
yes getting rid of Compact is fine.

tknopp:
thanks!

timholy:
Hmm, it's still using macOS 10.13 despite my efforts to change that. I suspect we should merge #438 if there are significant documentation improvements.

tknopp:
yes lets do that. With BB on the horizon, we should give up on Mac for the transition time.

PR #444, opened 2019-09-23
Update docs

remi-garcia:
As @jonathanBieler pointed out in https://github.com/JuliaGraphics/Gtk.jl/pull/438#issuecomment-534022823 build passes but docs update fails.

I think `.travis.yml` needs a cleanup. Documenter.jl was pinned to v0.19

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: julia
+coveralls: true
 sudo: required
 os:
   - linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,8 @@ addons:
       - xvfb
       - xauth
       - libgtk-3-dev
+script:
+  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
 jobs:
   include:
     - stage: "Documentation"
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,17 +2,17 @@ language: julia
 sudo: required
 os:
   - linux
-  #- osx
+  - osx
 julia:
   - 0.7
   - 1.0
   - 1.2
   - nightly
-# uncomment the following lines to allow failures on nightly julia
 # (tests will run but not make your overall status red)
 matrix:
   allow_failures:
     - julia: nightly
+    - os: osx
 notifications:
   email: false
 git:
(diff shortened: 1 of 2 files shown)


remi-garcia:
This doesn't solve docs update fails. My understanding is that we need a Member of JuliaGraphics to toggle some parameters on travis-ci.org.

.travis.yml
```diff
@@ -27,20 +27,12 @@ script:
...
-                 cd(joinpath(dirname(pathof(Gtk))));
-                 ENV["DOCUMENTER_DEBUG"] = "true";
-                 include(joinpath("docs", "make.jl"))'
+jobs:
+  include:
+    - stage: "Documentation"
+      julia: 1.0
+      os: linux
+      script:
+        - julia --project=docs/ -e 'import Pkg; Pkg.instantiate(); Pkg.develop(Pkg.PackageSpec(path=pwd()))'
```
vtjnash:
```suggestion
        - if [[ `uname` = "Linux" ]]; then TESTCMD="xvfb-run julia"; else TESTCMD="julia"; fi
        - $TESTCMD --project=docs/ -e 'import Pkg; Pkg.instantiate(); Pkg.develop(Pkg.PackageSpec(path=pwd()))'
        - $TESTCMD --project=docs/ docs/make.jl
```

remi-garcia: To build doc we precise to use linux. Is that `$TESTCMD` really needed?

tknopp: I don't think so. @vtjnash what where the intentions?

tknopp:
@remi-garcia: The documentation stage on travis is still red with this PR

remi-garcia:
I think we need a member of JuliaGraphics. I think we have an issue with [SSH keys](https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#SSH-Deploy-Keys-1) on travis-ci.org
I don't know who to ping, none of them are watching this repo.

tknopp:
No travis reports an issue with Mir: https://travis-ci.org/JuliaGraphics/Gtk.jl/jobs/588499948

remi-garcia:
That error can't occur because of a wrong ssh key?
```Failed to connect to Mir: Failed to connect to server socket: No such file or directory
Unable to init server: Could not connect: Connection refused```

Tbh I've never configure travis all by myself yet. Sorry if that's naïve.

tknopp:
fixed in #450

PR #445, opened 2019-09-24
More text methods

jonathanBieler:
I've added a couple of text related methods (search, place_cursor, select_range, selection_bounds, create_mark) and more options for the `skip` methods. 

I also added docstrings to the methods here, so we can start to move toward a more comprehensive documentation.


tknopp:
looks good.

PR #447, opened 2019-10-03
Initial shot at using Gtk on Julia 1.3+

staticfloat:
Note that this requires the Julia 1.3+ version of `cairo` as well (which is why the `Project.toml` shows `Cairo.jl` as being dev'ed)

~This doesn't work totally yet; it crashes halfway through the tests on my mac because it can't load some icons.~

This now passes all tests locally, but the windows that pop up are never drawn into; they remain white.  More debugging information below.

PR #448, opened 2019-10-09
`_jl_libcairo` -> `libcairo`

asinghvi17:


tknopp:
what's the purpose of this change? Gtk.jl is not compatible with Cairo.jl v0.7.

giordano:
Note also that this is already in #447

PR #451, opened 2019-10-13
remove Gtk2 code

tknopp:
Given that this package has not too much resources working on it, I think it is better to drop Gtk2 code.


jonathanBieler:
Seems good to me.

tknopp:
Ok, good. I know that we had some Gtk2 user at some point but it now seems not to be justified anymore.

jonathanBieler:
Plus getting binaries for Gtk3 is already hard enough as is.

Issue #325, opened 2017-09-08
Gtk causes lags in REPL on Windows


sebastianpech:
The moment I do `using Gtk` in the REPL on Windows (Tested on 10 and 7 with julia 0.6) the REPL input lags (see screencapture). This doesn't happen (at least for me) on OSX and Linux. 
<img src="http://i.imgur.com/ZzuNEi5.gif" />

tknopp:
Task scheduler issue? ping @vtjnash

vtjnash:
Win32 kernel bug

tknopp:
is this specific to Gtk.jl or is the task machinery in general broken on Windows? Is there a dedicated issue in the julia issue list?

hustf:
Very few people use win32 nowadays. Win64 is fast now. But we might try and see if processor optimizations has benefits on win64. I will try to compile the libraries when I get to it. There is also some workaround for julia itself, but last times I tried the gtk binaries still had issues.

davidanthoff:
This is actually still very much an issue on Windows. I've been entirely confused why the REPL experience on Windows seems to degenerate completely in certain situations, and it does just seem to be enough to load Gtk.jl.

It would be great if this could be a high priority issue to fix. There are a fair number of packages that load Gtk.jl, so a lot of folks will probably end up with Gtk.jl loaded (as I did), and then the REPL is really, really crappy.

SimonDanisch:
I can confirm! I had to remove Gtk from my project since it was unbearable...

tknopp:
@vtjnash could you elaborate what you mean by "Win32 Kernel Bug" above? Gtk.jl once worked fine on windows and now it seems that people have issues.

hustf:
GTK may have worked fine on windows at one time, but I have been trying it out now and then since mid 2015 and always gave up. Tests pass most of the time, but there is always some problem or other. 

I give up easily. GTK didn't make all binaries available for Windows in the past, and GTK.jl didn't test for windows. I'm reluctant to use anything that can't ultimately be used in my line of work, i.e. the Windows world. And I'll never be able to compile anything at work.

JuliaGraphics has an Appveyor account, which is used [here](JuliaGraphics/ColorSchemes.jl): That means that Simon or some other full member can add this project, if it's not already there. 

[...]

Issue #181, opened 2015-09-22
Strange interaction with redirect_stdout


jonathanBieler:
I'm trying to display Julia output in a GtkWidget but it's kind of slow and feels sluggish. I put together this minimal example (coming from IJulia):

    function send_stream(rd::IO, name::AbstractString)
        nb = nb_available(rd)
        if nb > 0
            d = readbytes(rd, nb)
            s = try
                bytestring(d)
            catch
            end
            #do something with s
        end
    end

    function watch_stream(rd::IO, name::AbstractString)
        try
            while !eof(rd) # blocks until something is available
                send_stream(rd, name)
                sleep(0.01) # a little delay to accumulate output
            end
        catch e
            # the IPython manager may send us a SIGINT if the user
            # chooses to interrupt the kernel; don't crash on this
            if isa(e, InterruptException)
                watch_stream(rd, name)
            else
                rethrow()
            end
        end
    end

    global read_stdout
    read_stdout, wr = redirect_stdout()
    function watch_stdio()
        @async watch_stream(read_stdout, "stdout")
    end
    watch_stdio()

Calling a print command (e.g. `print(randn(100))`) in the REPL is very fast (after compilation). However as soon as you use Gtk (`using Gtk`) it becomes very slow (seems like 2 seconds on my computer). I've tried with a few other packages but Gtk seems to be the only one interacting.

Any idea what's going on and how to fix that?

vtjnash:
what platform? I see no difference on OSX.

jonathanBieler:
Windows 7.

I've been commenting out parts of Gtk.jl and the problem appears as soon as `events.jl` get included.

vtjnash:
what happens if you change the `100` at https://github.com/JuliaLang/Gtk.jl/blob/49f0cc2fb2225d91f96abe0467cfba1ec87dfc6a/src/GLib/signals.jl#L358 to something smaller, like `10`?

jonathanBieler:
It does decrease the time a bit (and allocates more), I went down to 1 which seems to be the fastest:

    @time println(rand(100))

    1:
    2.011359 seconds (3.64 M allocations: 55.695 MB, 0.11% gc time)
    10:
    2.047513 seconds (1.41 M allocations: 22.999 MB, 0.17% gc time)
    100:
    2.352942 seconds (939.67 k allocations: 14.448 MB, 0.05% gc time)

[...]

PR #454, opened 2019-11-26
Usage of Gtk on windows needs a warning

SimonDanisch:
After the second time wasting quite a bit of time to debug why everything is suddenly sluggish, until I noticed that some unexpected third party library was loading in Gtk, I come to the conclusion that we really need this warning.

### DIFF
--- a/src/events.jl
+++ b/src/events.jl
@@ -12,7 +12,10 @@ function __init__()
             (Ptr{Nothing}, Ptr{Nothing}, Ptr{UInt8}, Ptr{Nothing}, Ptr{UInt8}, Ptr{GError}),
             C_NULL, C_NULL, "Julia Gtk Bindings", C_NULL, C_NULL, error_check)
     end
-
+    
+    if Sys.iswindows()
+        @warn "You are using Gtk on Windows which is currently buggy. Expect your REPL/IDE and anything depending on task switches become sluggish."
+    end
     # if g_main_depth > 0, a glib main-loop is already running,
     # so we don't need to start a new one
     if ccall((:g_main_depth, GLib.libglib), Cint, ()) == 0


giordano:
I think this should be fixed in #447, if I remember correctly by https://github.com/JuliaGraphics/Gtk.jl/pull/447/files#diff-8c2b506e7179e52e816713686d731068 in particular

SimonDanisch:
That'd be awesome! Not really a reason against merging this, I guess ;) 
Once #447 lands and removes the problem we can get rid of the warning!

SimonDanisch:
Oh, bad news Btw...From all I can see, I'm using #447, so doesn't seem fixed...

davidanthoff:
Yes, very much in favor of merging this, I've been bitten too many times by this!

jonathanBieler:
I'm fine with merging this.

tknopp:
I am not so happy with the warning, but its the peoples choice who use windows. @jonathanBieler are you using Windows? Did the fix of @vtjnash really did not fix the issue? On OSX it did.

The plan now is as follows:
- Lets put upper bounds on all depending packages and make a release
- Then merge the JLL branch
- Wait for all depending JLL branches (Cairo!) to be released
- Release a new release for Julia >= 1.3

giordano:
Maybe merge Cairo jll pull request before the one for gtk?

tknopp:
Thats what I said with the third point, no?

tknopp:
I mean for 2/3 it does not really matter to much, what the order is.

jonathanBieler:
I haven't used Gtk on windows for a while (I could give it a try) but as I understand the underlying issue has never been tackled (or maybe I missed it). #325 #181

tknopp:
The "fix" is in the JLL branch, see this:
https://github.com/JuliaGraphics/Gtk.jl/pull/447/files#diff-8c2b506e7179e52e816713686d731068L353

The communication is within the binarybuilder channel of slack. This was done by Jameson and Elliot.

tknopp:
note that I needed dropping this as it conflicted badly with the JLL branch that has priority. The function where this was integrated was not present in the JLL branch :-)

PR #460, opened 2019-12-05
Support Graphics 1.0 and test set_coordinates (+Documenter bounds)

timholy:
In preparation for https://github.com/JuliaGraphics/Graphics.jl/pull/16. It's better if the `set_coordinates` test gets done here since this the top of the dependency hierarchy. I have a suspicion it will fail based on https://github.com/JuliaGizmos/GtkReactive.jl/pull/91 (and that this is a regression), but we will see.


timholy:
Yep, as I figured, this is a regression compared to the pre-JLL days.  The warning about "infinite surface size not supported" is new; this might or might not be related to the failure to set the transformation (https://www.cairographics.org/manual/cairo-Transformations.html).

PR #465, opened 2019-12-09
Add a few precompiles

timholy:
Inference is not a major source of slowdowns for this package, but it doesn't hurt to add these.


tknopp:
is this snooped? And if not, would it be better to do so? (your are the expert)

timholy:
Yes, it was generated that way. I tend to edit the resulting precompile files though, and sometimes craft sections by hand.

With `@snoopi` only the "trigger" for inference is shown. Any dependent methods that are known by inference to be called also get precompiled, without you having to specify those. It makes the precompile files much shorter, which is nice.

tknopp:
Thanks for the explanation, then this seems to be ready to go.

PR #480, opened 2020-01-08
added new example

NandVinchhi:
This is the Gtk.jl implementation of a simple calculator.


logankilpatrick:
Nice work here @NandVinchhi 

see https://www.youtube.com/watch?v=9-KJKEbttJo&loop=0 for info on this code

PR #501, opened 2020-04-14
Allow specifying current folder and suggesting filenames in file dialogs

halleysfifthinc:
- `open_dialog` only supports `current_folder`
  - Suggesting a filename to open seems unlikely to be needed or helpful, but we could still
  support setting `current_name` if that seems like a more thorough support of these features.
- `save_dialog` supports both new kwargs

Gtk documentation [recommends against](https://developer.gnome.org/gtk3/stable/GtkFileChooserDialog.html#gtkfilechooserdialog-setting-up)
setting the current folder due to some heuristics in `GtkFileChooser` which are supposed to set sensible/helpful
folders, in my experience the dialogs only ever opened to the "Recent" folder.

I tested that all of the combinations of function & kwargs works properly; it doesn't seem like there is any
other way to test these features via code?

I also added examples to the documentation.

Refs: 
https://developer.gnome.org/gtk3/stable/GtkFileChooser.html#gtk-file-chooser-set-current-name 
https://developer.gnome.org/gtk3/stable/GtkFileChooser.html#gtk-file-chooser-set-current-folder

### DIFF
--- a/docs/src/manual/filedialogs.md
+++ b/docs/src/manual/filedialogs.md
@@ -33,9 +33,14 @@ if isdir(dir)
 end
 ```
 
+The open and save dialogs (incl. native versions) support the keyword argument `current_folder`
+to set the current folder shown when the dialog opens. Both save dialog variants support an additional
+keyword argument, `current_name`, to allow suggesting a filename for saving.
+
 Here are some examples:
 ```julia
 open_dialog("Pick a file")
+open_dialog("Pick a file"; current_folder=homedir())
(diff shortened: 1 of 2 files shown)


tknopp:
Great, could you still add a test, where at least the code is run? In that way at least we can guarantee that the code is runnable.

bilderbuchi:
This seems useful, I was _just_ looking for how to set this. Any chance it can be merged?

halleysfifthinc:
:+1: I'll try to get back to this sometime this week.

PR #504, opened 2020-04-26
Create an API reference page for currently undocumented methods

asinghvi17:


tknopp:
I actually don't have a clue what this will look like. Will merge to see.

PR #512, opened 2020-06-03
fix for julia v1.5; `jl_function_ptr` was removed

JeffBezanson:
This seems to work well enough for now as a workaround; ImageView.jl works at least. Would be good to try some more Gtk apps to confirm whether it works. I wanted to to just un-break 1.5, and we can work on a redesign later.

### DIFF
--- a/src/events.jl
+++ b/src/events.jl
@@ -35,7 +35,7 @@ end
 
 mutable struct Gtk_signal_motion{T}
     closure::T
-    callback::Ptr{Nothing}
+    callback::Function
     include::UInt32
     exclude::UInt32
 end
@@ -43,11 +43,7 @@ function notify_motion(p::Ptr{GObject}, eventp::Ptr{GdkEventMotion}, closure::Gt
     event = unsafe_load(eventp)
     if event.state & closure.include == closure.include &&
        event.state & closure.exclude == 0
-        if isbitstype(T)
-            ret = ccall(closure.callback, Cint, (Ptr{GObject}, Ptr{GdkEventMotion}, T), p, eventp, closure.closure)
-        else
-            ret = ccall(closure.callback, Cint, (Ptr{GObject}, Ptr{GdkEventMotion}, Any), p, eventp, closure.closure)
-        end
+        ret = closure.callback(p, eventp, closure.closure)
     else
         ret = Int32(false)
     end
@@ -70,14 +66,8 @@ function on_signal_motion(move_cb::Function, widget::GtkWidget,
         mask |= GdkEventMask.BUTTON_MOTION
     end
     add_events(widget, mask)
-    @assert Base.isstructtype(T)
-    if isbitstype(T)
-        cb = cfunction_(move_cb, Cint, (Ptr{GObject}, Ptr{GdkEventMotion}, T))
-    else
-        cb = cfunction_(move_cb, Cint, (Ptr{GObject}, Ptr{GdkEventMotion}, Ref{T}))
-    end
     closure = Gtk_signal_motion{T}(
-        closure, cb,
+        closure, move_cb,
         UInt32(include),
         UInt32(exclude)
         )
--- a/src/GLib/GLib.jl
+++ b/src/GLib/GLib.jl
@@ -30,9 +30,11 @@ export @sigatom, cfunction_
 
 
 cfunction_(f, r, a::Tuple) = cfunction_(f, r, Tuple{a...})
-@noinline function cfunction_(f, r, a)
-    @nospecialize(f, r, a)
-    return ccall((:jl_function_ptr,:libjulia), Ptr{Cvoid}, (Any, Any, Any), f, r, a)
+
+@generated function cfunction_(f, R::Type{rt}, A::Type{at}) where {rt, at<:Tuple}
+    quote
+        @cfunction($(Expr(:$,:f)), $rt, ($(at.parameters...),))
+    end
 end
 
 # local function, handles Symbol and makes UTF8-strings easier


src/events.jl
```diff
@@ -35,19 +35,15 @@ end
 
 mutable struct Gtk_signal_motion{T}
     closure::T
-    callback::Ptr{Nothing}
+    callback::Base.CFunction
```
vtjnash, line 38:
```suggestion
    callback
```

src/events.jl
```diff
@@ -35,19 +35,15 @@ end
...
 function notify_motion(p::Ptr{GObject}, eventp::Ptr{GdkEventMotion}, closure::Gtk_signal_motion{T}) where T
     event = unsafe_load(eventp)
     if event.state & closure.include == closure.include &&
        event.state & closure.exclude == 0
-        if isbitstype(T)
-            ret = ccall(closure.callback, Cint, (Ptr{GObject}, Ptr{GdkEventMotion}, T), p, eventp, closure.closure)
-        else
-            ret = ccall(closure.callback, Cint, (Ptr{GObject}, Ptr{GdkEventMotion}, Any), p, eventp, closure.closure)
-        end
+        ret = closure.callback.f(p, eventp, closure.closure)
```
vtjnash, line 46:
```suggestion
        ret = closure.callback(p, eventp, closure.closure)
```

vtjnash: This would be preferable to avoid increasing yet more the dependence on internal design details.

timholy:
With ProfileView I get a long inference stacktrace from @JeffBezanson's version, and with @vtjnash's I get
```julia
julia> FATAL ERROR: Gtk state corrupted by error thrown in a callback:
ERROR: MethodError: objects of type Base.CFunction are not callable
Stacktrace:
 [1] notify_motion(::Ptr{GObject}, ::Ptr{Gtk.GdkEventMotion}, ::Gtk.Gtk_signal_motion{GtkReactive.MouseHandler{GtkReactive.UserUnit}}) at /home/tim/.julia/dev/Gtk/src/events.jl:46
 [2] (::Gtk.var"#237#238")() at /home/tim/.julia/dev/Gtk/src/events.jl:2
 [3] g_sigatom(::Any) at /home/tim/.julia/dev/Gtk/src/GLib/signals.jl:176
 [4] gtk_main() at /home/tim/.julia/dev/Gtk/src/events.jl:1
```

src/events.jl
JeffBezanson: I agree. I don't think we have a call overload for `CFunction` though.

vtjnash: No, but let's just not allocate it then

JeffBezanson:
Looking into it.

JeffBezanson:
Works on 1.5 --- the inference issue is due to a recent change on master.

tknopp:
Thanks a lot for having a look into this. Especially nice is that we do not need to change the public interface. Would be interesting how "bad" you would consider this workaround. As far a I can see you are using a generated function to move the call to `@function` into an earlier stage of the compiler. I know that generated functions should not be used too much, but in my opinion we currently only have a hand full calls to the native `signal_connect` function.

JeffBezanson:
I don't think it's that bad. Certainly better than calling an internal C function, so for purposes of keeping the current API it's fine.

timholy:
Do you want this merged as-is or do you want https://github.com/JuliaGraphics/Gtk.jl/pull/512#discussion_r434809688 and/or other changes?

giordano:
Note that arm64 is failing on Drone because tests that were broken aren't anymore:
```
SetCoordinates: Error During Test at /drone/src/test/gui.jl:548
 Unexpected Pass
 Expression: mtrx.xx == 300
 Got correct result, please change to @test if no longer broken.

SetCoordinates: Error During Test at /drone/src/test/gui.jl:549
 Unexpected Pass
 Expression: mtrx.yy == 280
 Got correct result, please change to @test if no longer broken.
```
armv7l is failing at init time instead:
```
Stacktrace:
 [1] pipeline_error at ./process.jl:525 [inlined]
 [2] read(::Cmd) at ./process.jl:412
 [3] (::Gtk.var"#301#307"{String})() at /drone/src/src/Gtk.jl:94
 [4] withenv(::Gtk.var"#301#307"{String}, ::Pair{String,String}) at ./env.jl:161
 [5] (::Gtk.var"#300#306")(::String) at /drone/src/src/Gtk.jl:93
 [6] (::gdk_pixbuf_jll.var"#8#9"{Gtk.var"#300#306"})() at /root/.julia/packages/gdk_pixbuf_jll/kzXJa/src/wrappers/arm-linux-gnueabihf.jl:50
 [7] withenv(::gdk_pixbuf_jll.var"#8#9"{Gtk.var"#300#306"}, ::Pair{String,String}, ::Vararg{Pair{String,String},N} where N) at ./env.jl:161
 [8] #gdk_pixbuf_query_loaders#7(::Bool, ::Bool, ::typeof(gdk_pixbuf_jll.gdk_pixbuf_query_loaders), ::Gtk.var"#300#306") at /root/.julia/packages/gdk_pixbuf_jll/kzXJa/src/wrappers/arm-linux-gnueabihf.jl:49
 [9] gdk_pixbuf_query_loaders at /root/.julia/packages/gdk_pixbuf_jll/kzXJa/src/wrappers/arm-linux-gnueabihf.jl:33 [inlined]
 [10] __init__() at /drone/src/src/Gtk.jl:92
 [11] _include_from_serialized(::String, ::Array{Any,1}) at ./loading.jl:692
 [12] _require_from_serialized(::String) at ./loading.jl:743
 [13] _require(::Base.PkgId) at ./loading.jl:1034
 [14] require(::Base.PkgId) at ./loading.jl:922
 [15] require(::Module, ::Symbol) at ./loading.jl:917
 [16] include at ./boot.jl:328 [inlined]
 [17] include_relative(::Module, ::String) at ./loading.jl:1105
 [18] include(::Module, ::String) at ./Base.jl:31
 [19] include(::String) at ./client.jl:424
 [20] top-level scope at none:6
during initialization of module Gtk
in expression starting at /drone/src/test/runtests.jl:2
```

src/events.jl
JeffBezanson: Oh, wait a minute, is this only called from julia? Why was it using `cfunction` in the first place? @tknopp any reason?

timholy:
@tknopp, with apologies to you I presumed to release 1.1.4, so it doesn't hold back progress on the Julia 1.5 release.

tknopp:
That is just fine. Wanted to have a look at this right now but am happy that things are already done.

PR #521, opened 2020-09-11
Mitigate slowdown by decreasing the timeout

timholy:
xref #467

### DIFF
--- a/src/GLib/signals.jl
+++ b/src/GLib/signals.jl
@@ -297,7 +297,7 @@ function uv_prepare(src::Ptr{Nothing}, timeout::Ptr{Cint})
     else
         ccall(:uv_update_time, Nothing, (Ptr{Nothing},), evt)
         tmout_ms = ccall(:uv_backend_timeout, Cint, (Ptr{Nothing},), evt)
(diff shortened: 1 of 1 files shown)


SimonDanisch:
On this branch, I can see no sluggishness anymore

timholy:
Test failures seem unrelated. OK to merge?

tknopp:
if this really fixes the problem on windows we should get rid of the warning as well

Cruor:
I'm still experiencing keyboard issues on Gtk v1.1.5 (which seems to include this change). Tested on both Julia 1.4 and 1.5, just to make sure. For our application it is still sluggish for 10-30 seconds, can't tell if there is any improvements sadly.

0x0ade:
I'm also experiencing keyboard proble`m`s with Gtk.jl v1.1.5 and v1.1.6 with Julia 1.5.3, 1.6.0-beta1 and 1.^6^.0-rc1. In addition to the sluggishness, I'm experiencing this when starting up a program using Gtk.jl:

aaaaaaaaaaaââââáààáàaaaaaaaaaaaaaaa

I've held down the A key while the program was starting up and as you can see, my system seems to ^receive ghost key presses like this one right as the sluggishness starts to kick in. I can provide a video recording of this behavior^i´f necessary.

PR #523, opened 2020-09-14
Get rid of the REPL slowdown on windows that was fixed by #741

tknopp:
@timholy @SimonDanisch


timholy (approved):
If we're confident it basically ends the problem, yes, definitely eliminate this.

timholy:
Will be good to have a new release once this is gone.

Issue #355, opened 2018-06-15
Blank minimize/maximize buttons on Windows 7 with Julia 0.6.3


pauljurczak:
Here is how the window with default parameters looks like:

![image](https://user-images.githubusercontent.com/4933032/41449773-4a956f22-7020-11e8-83e2-722edf8b69e7.png)

tknopp:
Then something happened to WinRPM and it packages. Here are the interesting lines:
https://github.com/JuliaGraphics/Gtk.jl/blob/master/deps/build.jl#L30

pauljurczak:
It's plausible, as I vaguely remember some warnings related to WinRPM build. I will try on another PC.

pauljurczak:
I have the same problem on Windows 10 PC. I have these packages installed:
7 required packages:
 - Compose                       0.6.0
 - GLVisualize                   0.7.0
 - Gtk                           0.14.0
 - IJulia                        1.8.0
 - ImageMagick                   0.5.1
 - ImageView                     0.6.0
 - OpenCL                        0.7.0

tknopp:
I would suggest running all `WinRPM` commands manually and trying to see where an error occurred. This includes the "schema" command and the end of the code that I quoted.

mwawra:
I just want to note, that the issue, I think, was already present with Julia v0.6.2.

To reproduce the problem one can also just load the gtk3-icon-browser from C:\Users\<username>\.julia\v0.6\WinRPM\deps\usr\x86_64-w64-mingw32\sys-root\mingw\bin. There you can also see that the symbolic-icons didn't work. I already installed gdk-pixbuf-loader-rsvg for svg-files but this didn't solved the problem. I think it really is a problem with the WinRPM-packages.

w1res:
I'm still getting this on julia 1.0.1 on Windows 10.

pauljurczak:
I'm also experiencing this behavior with a fresh install of Julia 1.0.1 on Windows 10.

pauljurczak:
Here is the solution I found: set environment variable `GTK_CSD=0` to enable Windows native decorations. It works fine with Windows 7 and 10. Thank you, stackoverflow!

tknopp:
Cool. Could you check if is possible to set that environment variable from within Julia? Then it could be set on Windows during initialization from Gtk.jl

pauljurczak:
[...]

PR #548, opened 2020-12-29
Fix #547 and revert #355

### LINKED ISSUE #547: Title bar buttons are now displayed properly on Windows with GTK_CSD=1, #355 should be reversed
Currently GTK_CSD is forced to be 0 on Windows because title bar buttons used to be broken previously. Now title bar buttons are fixed and displayed properly - using the Ahorn Julia package as an example:
image
So, it should be allowed to set GTK_CSD to 1 on Windows. This can be fixed by deleting these lines from src/Gtk.jl:
https://github.com/JuliaGraphics/Gtk.jl/blob/f87935578d9610ca15ee5efd8eaf7cdec7775b9a/src/Gtk.jl#L112-L115

tknopp:
With the new BB binaries the window decorations show correctly without setting GTK_CSD=1

close #547 

@ping NeoKat75: I cannot test this, have you tested it with exactly this change?

### DIFF
--- a/src/Gtk.jl
+++ b/src/Gtk.jl
@@ -109,11 +109,6 @@ function __init__()
         )
     end
 
-    if Sys.iswindows()
-      # needed on windows for correct window decorations (issue 355)
-      ENV["GTK_CSD"] = 0
-    end
-
     # Point gdk to our cached loaders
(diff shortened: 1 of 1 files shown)


NeoKat75:
> @NeoKat75: I cannot test this, have you tested it with exactly this change?

@tknopp yep, it works just fine!

PR #549, opened 2020-12-30
Slightly expand precompiles

timholy:
These were found via the new capabilities in the soon-to-be-released SnoopCompile. Not a big change but you can see the time savings in the comment after each line.


tknopp:
Nice!

PR #551, opened 2020-12-31
Yet more precompiles

timholy:
These were identified similarly to #549


timholy:
These forthcoming tools in SnoopCompile are awesome. Time for ProfileView to produce a flamegraph window:
- master: 2.63s
- upcoming release: 0.94s

There will be a lot of updated packages before that becomes widely available (including a new GtkObservables package).

PR #552, opened 2021-01-02
Reduce specialization in GLib methods

timholy:
A lot of these methods run on non-inferrable argument types.
Consequently it seems to make sense to reduce the amount of
specialization.

Note there's one case with *increased* specialization, to make it
more reliably inferrable and thus precompilable.

I *think* all of these make sense, but if anyone can review it I would certainly appreciate it.


timholy:
I'll merge soon barring objections.

tknopp:
go ahead, I cannot really review that because you are far ahead in knowledge about all this specialization stuff.

PR #555, opened 2021-01-14
Add environment for generator scripts

Vexatos:
Minimal Project file for the generator scripts. See https://github.com/JuliaGraphics/Gtk.jl/pull/535#issuecomment-760067299.

From CI or tests, it'd be called like this:

```julia
using Gtk # latest Gtk from dev, imported in the main or test Gtk env
using Pkg
Pkg.activate("$root/gen")
Pkg.instantiate()
include("gtk_auto_gen.jl")
```
The generator scripts depend on the dev version of Gtk.jl, but actually making the environment depend on it isn't possible without including the `Manifest.toml` (using `repo-url = ".."`), which probably shouldn't be done.

### DIFF
--- a/gen/Project.toml
+++ b/gen/Project.toml
@@ -1,0 +1,4 @@
+[deps]
+Clang = "40e3b903-d033-50b4-a0cc-940c62c95e31"
+GTK3_jll = "77ec8976-b24b-556a-a1bf-49a033a670a6"
+Serialization = "9e88b42a-f829-5b0c-bbe9-9e923198166b"


tknopp:
But aren't the tests run in dev mode? Not sure how this is accomplished in the test's project file

Vexatos:
Testing in julia is a kind of special environment, so yes, it does, meaning if you were to run it from inside the test environment (not that I'd recommend that) you'd get the correct output. Similarly, you can run it starting from the primary Project.toml in any CI step for the same effect.

PR #558, opened 2021-02-05
Added miscellaneous text and window related bindings

jonathanBieler:
I've added the bindings for a couple of functions that are not in GAccessor. I think that's the last batch of methods I had in my GtkExtensions package.

I haven't had the time to add tests so I didn't exported the methods (not sure if we want to export everything all the time), but I think it's still good to have those in here. I'll add some if I have the opportunity. I also fixed a couple of formatting issues.


jonathanBieler:
If no one objects I'll merge this later.

Issue #562, opened 2021-02-12
cascading failure causing segfault in GtkReactive tests?


vtjnash:
https://s3.amazonaws.com/julialang-reports/nanosoldier/pkgeval/by_date/2021-02/11/GtkReactive.1.7.0-DEV-50742c7e42.log
```
(julia:170): GLib-GObject-CRITICAL **: 06:21:42.833: g_object_ref_sink: assertion 'G_IS_OBJECT (object)' failed

(julia:170): GLib-GObject-CRITICAL **: 06:21:42.843: g_object_setv: assertion 'G_IS_OBJECT (object)' failed

(julia:170): GLib-GObject-CRITICAL **: 06:21:42.872: g_closure_new_object: assertion 'G_IS_OBJECT (object)' failed

signal (11): Segmentation fault
in expression starting at /home/pkgeval/.julia/packages/GtkReactive/gQSXq/test/runtests.jl:22
unsafe_store! at ./pointer.jl:118 [inlined]
_signal_connect at /home/pkgeval/.julia/packages/Gtk/TgEIW/src/GLib/signals.jl:40
signal_connect at /home/pkgeval/.julia/packages/Gtk/TgEIW/src/GLib/signals.jl:27 [inlined]
signal_connect at /home/pkgeval/.julia/packages/Gtk/TgEIW/src/GLib/signals.jl:27
```

tknopp:
Is this a) reproducible and b) also happens in 1.6? I also have seen segfaults on 1.6-beta on an internal software of mine. A minimal reproducible example would be good to track this down.

giordano:
Those are the tests run by PkgEval. Note that tests for nightly have been been failing forever since we switched to Github Actions: https://github.com/JuliaGraphics/Gtk.jl/actions?query=workflow%3ACI

tknopp:
Hm, yes. But nightly != 1.6 right?

giordano:
Yes, but I was referring to point 1, tests have been consistently failing for months. Maybe looking at them might help. Julia v1.6 isn't tested at the moment in CI (but you just need to add one line to the matrix)

tknopp:
Hm, "Error: Could not find a Julia version that matches 1.6". Do you know what to enter there? 1.6-?

tknopp:
tests passing locally on Julia 1.6rc1

giordano:
[...]

PR #565, opened 2021-02-23
Fix tests on 1.6

timholy:
It seems to need a little delay before checking `:active`.


giordano:
It doesn't seem to be always enough (did Julia get too fast? :sweat_smile:).

Good news is that this makes CI more happy, bad news is that CI failure is probably not related to #562?

timholy:
Yeah, it's weird. I suspect we can fix it by adding sufficient delay but that seems like the wrong way to go about it. Let's see what more knowledgeable folks say about https://github.com/JuliaGraphics/Gtk.jl/issues/562#issuecomment-784224932

timholy:
Superseded by #569 and #600

PR #571, opened 2021-05-07
[CI] Remove redundant version of Julia from matrix

giordano:
I'm using this PR also to test that the package works with GTK 3.24.29, but will remove that change afterwards


giordano:
Ok, good news, all tests on GitHub Actions are successful with GTK 3.24.29.  I'm going to remove the installation of `GTK3_jll#main`

Issue #274, opened 2016-12-30
API Unification


tknopp:
I think it is pretty complicated that Gtk.jl provides different APIs (regular and ShortNames). Have not seen this in any Julia package elsewhere. 

Whats you opinion on that @timholy @vtjnash. Was there a specific motivation for this split in API.

I don't know what my actual proposal is, but to be consistent what other packages are doing probably the best would be making the ShortNames the default. One issue is that Julia base/core has a type `Box` that would conflict with the Gtk version of that.

lobingera:
```
               _
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: http://docs.julialang.org
   _ _   _| |_  __ _   |  Type "?help" for help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.5.1-pre+2 (2016-09-20 03:34 UTC)
 _/ |\__'_|_|_|\__'_|  |  Commit f0d40ec* (103 days old release-0.5)
|__/                   |  x86_64-linux-gnu

julia> Box(4)
ERROR: UndefVarError: Box not defined

julia> box(4)
ERROR: UndefVarError: box not defined

julia> Base.box
(intrinsic function #0)

julia> Core.box
ERROR: UndefVarError: box not defined

julia> Base.Box
ERROR: UndefVarError: Box not defined

julia> Core.Box
Core.Box

```

so it looks like they are not exported?

timholy:
I wasn't involved until after that choice was made, but my impression was that it was to increase familiarity for people who come from a Gtk background. I agree with your general point, however, and I suspect scoping + short names is the way to go. @vtjnash may have a different perspective, however.

tknopp:
@lobingera: It seems you are right.

lobingera:
... i don't remember the reference (could be a github issue or a julia-users discussion):
use orginial names without the library prefix and prefix with julia module name: gtk_window_new -> Gtk.window_new ?

In any case it should be a notation which has a 1:1 mapping back to the Gtk c-API for easier lookup.

And generally i'd like to see Gtk.jl to be split into smaller, independent parts, like a Glib and Gdk module/package, so the library dependencies can be followed easier (Glib is not only a dependency for Gtk, but also for Cairo and Pango and...).

tknopp:
[...]

PR #591, opened 2021-11-02
ComboBoxText: Add methods for setting/getting active index/text

BioTurboNick:
Provides for direct access to the index of the selected item and the selected text.

### DIFF
--- a/src/basic_exports.jl
+++ b/src/basic_exports.jl
@@ -11,7 +11,7 @@ export new, width, height, #minsize, maxsize
     convert_iter_to_child_iter, convert_child_iter_to_iter,
     pulse,
     buffer, cells, search, place_cursor, select_range, selection_bounds,
-    create_mark, scroll_to
+    create_mark, scroll_to, setactiveindex!, getactiveindex, getactivetext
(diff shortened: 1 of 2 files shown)


tknopp:
isn't this captured by `GAccessor.active_text` and `Gtk.GAccessor.active`? Whenever we can we use the automatically generated accessors.

BioTurboNick:
It could be, it's not well-documented though.

BioTurboNick:
Ah, I think I see the intent. Nevermind then.

tknopp:
yes, Gtk.jl could be much better documented. I did in the past exactly what you did until Jameson told me to use the GAccessor module.

BioTurboNick:
Although, one issue is that `GAccessor.active_text()` can return a null pointer. Should this package be directly exposing that sort of thing, or Julia-izing it? (0-based to 1-based indexes, NULL => Nothing, etc.)

I see this has been raised before: https://github.com/JuliaGraphics/Gtk.jl/issues/274

tknopp:
Absolutely, the current stage is not perfect. There is room for lots of improvements.

PR #595, opened 2021-11-16
Fix incorrect method leading to silent exceptions in signal handlers

### LINKED ISSUE #593: Exceptions being swallowed in signal handlers
For some reason, exceptions inside signal handlers are being swallowed.

The try/catch block in `function g_siginterruptible(f::Base.Callable, @nospecialize(cb))` catches it, but the `@async` display code never prints to the REPL.

BioTurboNick:
Resolves #593


tknopp:
thanks!

PR #598, opened 2021-11-26
Avoid full CI on changes to docs

JeffFessler:
I noticed that #597 was doing a full CI run even though it is just a change to a `.md` file.
This (untested) PR hopefully should reduce the carbon footprint of future PRs that are just documentation updates.
It's quite possible that it needs tweaking to be correct.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,8 @@ on:
     branches: "master"
     tags: ["*"]
   pull_request:
+    paths-ignore:
+      - '**.md'
   release:
 
 jobs:


giordano:
> This (untested) PR

Why not doing the change in the same PR so you'd actually test it?

tknopp:
I cannot see a difference

JeffFessler:
> Why not doing the change in the same PR so you'd actually test it?

I was trying to follow this guidance:
https://github.com/SciML/ColPrac#guidance-on-contributing-prs
"PRs should do a single thing, so that they are easier to review."

Also, I somehow doubt that proposed changes to CI files affect the CI of the PR; presumably the CI files in the main branch govern the CI, not those in the PR?

I can't find the original post where I learned about this, but the basic idea is here:
https://michaelcurrin.github.io/dev-cheatsheets/cheatsheets/ci-cd/github-actions/triggers.html

PR #600, opened 2021-12-06
Preserve GValues from GC (fixes #581)

### LINKED ISSUE #581: New segfaults on nightly
It doesn't happen every time (see https://github.com/JuliaGizmos/GtkObservables.jl/pull/8 for full history), but it happened once on Linux and once on Mac, only on nightly. The pattern of segfaulting is the same for both:
```
signal (11): Segmentation fault
in expression starting at /home/runner/work/GtkObservables.jl/GtkObservables.jl/test/runtests.jl:547
type_get_qdata_L at /workspace/srcdir/glib-2.68.1/build_glib/../gobject/gtype.c:3702 [inlined]
type_check_is_value_type_U at /workspace/srcdir/glib-2.68.1/build_glib/../gobject/gtype.c:4222 [inlined]
g_type_check_value_holds at /workspace/srcdir/glib-2.68.1/build_glib/../gobject/gtype.c:4262
g_value_set_boolean at /workspace/srcdir/glib-2.68.1/build_glib/../gobject/gvaluetypes.c:745
_gtk_boolean_handled_accumulator at /workspace/srcdir/gtk+-3.24.29/build-gtk/../gtk/gtkprivate.c:139
accumulate at /workspace/srcdir/glib-2.68.1/build_glib/../gobject/gsignal.c:3241 [inlined]
signal_emit_unlocked_R at /workspace/srcdir/glib-2.68.1/build_glib/../gobject/gsignal.c:3746
g_signal_emitv at /workspace/srcdir/glib-2.68.1/build_glib/../gobject/gsignal.c:3227
signal_emit at /home/runner/.julia/packages/Gtk/P4Gsu/src/GLib/signals.jl:145
unknown function (ip: 0x7fd62410aa00)
_jl_invoke at /buildworker/worker/package_linux64/build/src/gf.c:2245 [inlined]
...
```
after which it has a long string of runtime (mostly interpreter).

https://github.com/JuliaGizmos/GtkObservables.jl/pull/8/checks?check_run_id=3102063969
https://github.com/JuliaGizmos/GtkObservables.jl/runs/3098639995

timholy:
As a consequence of the forced specialization on the `RT` argument
of `signal_emit` in #552, the compiler now knows whether
`RT === Nothing`. In that case, it also knows that
`return_value` will not be used, so it is free to be garbage-collected.
When that happens, it triggers segfaults. EDIT: actually, that can't
be correct, because it doesn't create them when `RT===Nothing`.
So probably #552 didn't trigger this.

This puts both potential `GValue`-arguments inside a `GC.@preserve`
to prevent garbage collection. Fixes #581.

A reproducer of the crash is with GtkObservables:
```sh
/some/path/GtkObservables$ julia --check-bounds=yes --color=yes --depwarn=yes --inline=yes --project=@. -e 'import Pkg; Pkg.test(; coverage=true)'
```
but that is obviously difficult to convert into a test. I've run it 5 times locally and have yet to see a crash, which did not happen before this commit.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Gtk"
 uuid = "4c0ca9eb-093a-5379-98c5-f87ac0bbbf44"
-version = "1.1.9"
+version = "1.1.10"
 
 [deps]
 Cairo = "159f3aea-2a34-519c-b102-8c37f9878175"
--- a/src/GLib/signals.jl
+++ b/src/GLib/signals.jl
@@ -142,7 +142,8 @@ function signal_emit(w::GObject, sig::AbstractStringLike, ::Type{RT}, args...) w
     end
     signal_id = ccall((:g_signal_lookup, libgobject), Cuint, (Ptr{UInt8}, Csize_t), sig, G_OBJECT_CLASS_TYPE(w))
     return_value = RT === Nothing ? C_NULL : gvalue(RT)
-    ccall((:g_signal_emitv, libgobject), Nothing, (Ptr{GValue}, Cuint, UInt32, Ptr{GValue}), gvalues(w, args...), signal_id, detail, return_value)
+    gvals = gvalues(w, args...)
+    GC.@preserve gvals return_value ccall((:g_signal_emitv, libgobject), Nothing, (Ptr{GValue}, Cuint, UInt32, Ptr{GValue}), gvals, signal_id, detail, return_value)
     RT === Nothing ? nothing : return_value[RT]
 end
 


timholy:
CC @BioTurboNick

timholy:
I ran the tests 20 more times and never got a segfault. I think we can safely say this fixes the problem that has been holding back GtkObservables.

tknopp:
then go ahead and press the button + release. It certainly does not hurt to prevent GC to collect that value during that call.

PR #602, opened 2021-12-20
get properties without specifying type

jwahlstrand:
As of GLib 2.60, g_object_get_property accepts an empty GValue, whose g_type gets set to that of the property's GParamSpec. This PR uses this in a get_gtk_property() method that doesn't take a type argument.

From there one could define `getproperty(w::GObject, name::Symbol)`, but that would collide with the "FieldRef" machinery. The approach taken here is to define `getindex(f::FieldRef)`, so that one could continue to use `t = w.title[String]` to access the "title" property as a String in a type stable way or `t = w.title[]` to access it in a type unstable way. It also defines `setproperty!(w::GObject)` which allows setting using `w.title = "My Title"` and `propertynames(w::GObject)` so that property names autocomplete in the REPL.

An alternative approach would remove the FieldRef stuff and rename get_gtk_property(), set_gtk_property!() to getproperty, setproperty!, and accept type instability for getting properties, allowing us to access properties by just typing `t = w.title`. It might be possible eventually to use GI to achieve type stability for getting properties -- I haven't figured that out.

I have a bit of experience using GTK in C, and I remember reading somewhere that using get_whatever() functions defined for most important properties is preferable performance-wise to using GObject properties, because it doesn't involve the GValue machinery. I've done a little benchmarking and that seems to be true, though the difference is not typically noticeable. So I think optimizing property handling for ease-of-use and recommending calls to set_whatever(), get_whatever() (i.e. GAccessor) when performance is a concern is the way to go.

If this change is accepted I would be happy to update the docs.


tknopp:
When I looked at this last, I was struggling a little bit the Gtk provides two ways for doing the same thing. It would be good if we could concentrate on the one. 

I think in the end we just want to have a single
```
t = w.title 
w.title = "My Title"
```
no matter what is being used under the hood. The first can always be explicitly casted if needed.

I have the following questions here:
- Do you know how the python Gtk bindings handle this?
- Are the sets of of getters/setter and get/setproperties in Gtk the same? Is one a subset of the other?

jwahlstrand:
I agree, simpler is better.

The python bindings define methods "get_property" and "set_property" and also GObjects have a props attribute so you can use `w.props.title = "My Title"`. https://pygobject.readthedocs.io/en/latest/guide/api/properties.html . It uses GI to generate all methods for GObjects, which include the getters and setters.

My impression is that the getters/setters are a subset of the properties but I will look into that.

tknopp:
But with your new GI approach we could do the same as python, no? Then we would automatically have type stability.

If the set available by the setter/getter is large enough to capture everything important, than I would somewhat favor that approach, since it is consistent with what the python bindings do.

jwahlstrand:
Sorry, I meant that I agree that keeping the property handling as simple and easy as possible is best: `t = w.title` is preferable to `t = w.title[]` which is preferable to `t = w.title[String]`. `t = w.props.title` is probably better than the last two too.

But for properties vs. getter/setters, I don't think it's an either/or issue -- both should be supported (like in the python bindings). The GI approach of exporting all methods will be equivalent to what's in GAccessor and look more like the python bindings. BTW libgirepository has functionality for connecting properties to their getter/setter methods, but it's not used by GTK3 -- unless I am completely missing something.

At this point I think I finally understand enough to start bringing GI in -- I'm hoping to have time in the next few weeks to propose a path forward. I thought I might as well propose this change to properties now since it doesn't involve GI.

tknopp:
Proposing the change now is much appreciated. But I hesitate merging this now if we know that we will change the interface soon again. Probably it would be better to generate a "v2" branch where you work on and gather there some major API changes that will in the end result in v2.0 of Gtk.jl?

I like `t = w.title` to be our primary interface and `t = w.props.title` to be the interface for the properties. So essentially, lets just do what Python does :-).

jwahlstrand:
OK, sounds good, I will keep plugging and resurface again later.

The initial interface may be more like `t = get_title(w)` and `t = w.props.title`...

tknopp:
And note that we don't need to overload v2 with too much features. I am all in favor for making a gradual transition to GI based stuff.

From the API perspective we should aim getting the property things in shape. And additionally IMHO we should get rid of `ShortNames` module. That is just confusing.

PR #605, opened 2022-01-02
use Librsvg_jll pixbuf loader (fixes #604)

### LINKED ISSUE #525: Cannot display SVG files, GtkSpinner not displayed
This is not necessarily a Gtk.jl issue, although this seems like the best place to talk about it.

Right now I can't open SVG images with Gtk.jl. If I try to display one, I get an error message due to the format not being recognized. It's a little more serious than that, though, because ability to display SVG is assumed in many places. For instance, I cannot display a GtkSpinner right now, and I suspect this is the reason.

Apparently Gtk usually relies on the rsvg library to render SVG files. Simply installing the library did not solve it for me. How do I enable Gtk.jl to display SVG?

I would be very happy with being able to display SVG. Failing that, it would be nice to know a workaround to make GtkSpinner use bitmaps.

### LINKED ISSUE #604: GtkSpinner doesn't appear.
```julia
using Gtk

win = GtkWindow("Spin Test")
spin = GtkSpinner()
push!(win, spin)
showall(win)

Gtk.start(spin)
```

The spinner doesn't show up. It also doesn't show up if I build the UI in Glade (where it works fine) and load the UI that way. Both Linux (via WSL) and Windows. Am I missing something?

jwahlstrand:
This uses Librsvg's pixbuf loader so that we can load SVG icons and render SVG through GdkPixbuf. It copies the loaders from gdk_pixbuf_jll and Librsbg_jll into a directory and runs gdk_pixbuf_query_loaders() on it. Thanks to @giordano for help with that last part.

Questions for the more highly educated than me: would it be better to create symbolic links rather than copy? What happens if Librsvg_jll or gdk_pixbuf_jll are updated? Is the MutableArtifacts.toml cache automatically regenerated after an update or should we handle that somehow?

Probably also fixes #525.


tknopp:
errors on Julia 1.3 due to some new `readdir` functionality being used. Can this  be implemented in a 1.3 compatible fashion?

jwahlstrand:
Yes, I'll fix that. I'll investigate the issue with Windows.

giordano:
Consider simply dropping support for Julia v1.5-, now that v1.6 is the LTS?

tknopp:
I am fine with that

jwahlstrand:
I can't figure out why Windows is failing, but it's occuring in `readchomp(addenv(...))`. I'll be able to return to this later.

jwahlstrand:
Not sure what to make of these CI failures.

The previous error on Windows went away by switching to withenv(), but GtkSpinner still doesn't work on Windows (it does on Linux and Mac). Now I'm seeing some permission errors (`g_module_open() failed for [artifact_dir\*.dll]: Access is denied`). However since my Windows computer is pretty broken it would be helpful if someone else could test this. @BioTurboNick ?

BioTurboNick:
@jwahlstrand Sure, checked it out.

Note - MutableArtifacts.toml has to be deleted to get the initializing code to rerun.

You might kick yourself for this, but:
```
return String(readchomp(`$gpql`))

to

return String(readchomp("$gpql"))
```

Gets rid of the errors, though the spinner still isn't working.

BioTurboNick:
Okay, here's something.

Other artifact directories are created with Full Control access permissions to the current user, and no permissions to Everyone.

Meanwhile, the special shared artifact directory created here has limited permissions for the current user (which is probably the most important part) and also adds Everyone.

image

BioTurboNick:
What I posted over in the Julia issue may be the problem. The original files have "Read & execute" permissions, but that permission gets overwritten by `Pkg`.

BioTurboNick:
Sadly, though, fixing the permissions manually doesn't seem to make the spinner appear.

BioTurboNick:
Ah, it works!

If I insert a pause between creating the shared artifact directory, set permissions to full control and override child permissions, then allow it to continue, I have a functioning spinner.

BioTurboNick:
@jwahlstrand - sorry I think I wasn't clear. I didn't mean to sleep the thread, I meant that if you manually set the permissions before it continues, it will then work.

jwahlstrand:
Oops, OK I'll remove that.

jwahlstrand:
It now works on my Windows computer (both 1.6 and 1.7). But in CI there are still failures for Windows: for Julia 1 (1.7) and nightly it's unable to find `Librsvg_jll.libpixbufloader_svg`.

jwahlstrand:
Ahh, I didn't notice at first that the issue is only on Windows x86. Now it's clear why -- Librsvg_jll is not built for that arch.

src/Gtk.jl
```diff
@@ -87,13 +89,26 @@ function __init__()
...
+    loaders_dir_name = "gdk-pixbuf-loaders-dir"
+    loaders_dir_hash = artifact_hash(loaders_dir_name, mutable_artifacts_toml)
     loaders_cache_name = "gdk-pixbuf-loaders-cache"
     loaders_cache_hash = artifact_hash(loaders_cache_name, mutable_artifacts_toml)
     if loaders_cache_hash === nothing
+        # Copy loaders into a directory
+        loaders_dir_hash = create_artifact() do art_dir
+            loaders_dir = mkdir(joinpath(art_dir,"loaders_dir"))
+            pixbuf_loaders = joinpath.(gdk_pixbuf_loaders_dir, readdir(gdk_pixbuf_loaders_dir))
+            isdefined(Librsvg_jll,:libpixbufloader_svg) && push!(pixbuf_loaders, Librsvg_jll.libpixbufloader_svg)
```
giordano, line 101:
```suggestion
            Librsvg_jll.is_available() && push!(pixbuf_loaders, Librsvg_jll.libpixbufloader_svg)
```

giordano:
> Now it's clear why -- Librsvg_jll is not built for that arch.

Because Rust toolchain for i686-w64-mingw32 is broken when using SJLJ unwinding (which is what we do): https://github.com/rust-lang/rust/issues/79609.  Judging by the comments of the developers, it's unlikely they'll fix it

src/Gtk.jl
jwahlstrand: Thanks!

tknopp:
ready? CI is passing.

jwahlstrand:
If this last commit passes, it's ready.

jwahlstrand:
I'll keep working on this. In the meantime I wouldn't let this delay a release.

jwahlstrand:
OK, the failure earlier must have been a fluke...I think this is good to go now.

Issue #503, opened 2020-04-22
Major slowdowns with Gtk + Threads


timholy:
This may be the same issue as #325, but observed on Linux with threads. I should say I'm running this on a 6-physical-core machine (12 if you include hyperthreads). Copied from https://github.com/JuliaImages/ImageFiltering.jl/issues/161 and specifically the test in https://github.com/timholy/ComputationalResources.jl/issues/18:

```julia
$ JULIA_NUM_THREADS=4 julia
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.4.2-pre.0 (2020-04-15)
 _/ |\__'_|_|_|\__'_|  |  release-1.4/ef4fe83698* (fork: 122 commits, 113 days)
|__/                   |

julia> using ImageFiltering, BenchmarkTools, ComputationalResources

julia> function test_imfilter(; opt=1)
               gk = Kernel.gaussian(2.0f0)
               I = rand(Float32,1000 ,1000)
               println("julia_num_threads=$(Threads.nthreads())")
               
               for i = 1:9
                       start = time()
                       if opt == 1
                               J = imfilter(CPU1(Algorithm.FIR()), I, gk)
                       elseif opt == 2
                               J = imfilter(I, gk)
                       else
                               error("Wrong opt value $opt, either 1 (for CPU1), 2 CPUThreads")
                       end
                       t = round(time() - start, digits=6)
                       println("$i imfilter $t sec")
               end
       end
test_imfilter (generic function with 1 method)

julia> test_imfilter(opt=1)
julia_num_threads=4
1 imfilter 0.020989 sec
2 imfilter 0.022114 sec
3 imfilter 0.035968 sec
4 imfilter 0.020871 sec
5 imfilter 0.020997 sec
6 imfilter 0.021044 sec
7 imfilter 0.024659 sec
8 imfilter 0.018204 sec
9 imfilter 0.018081 sec

julia> test_imfilter(opt=2)
julia_num_threads=4
1 imfilter 0.320837 sec
2 imfilter 0.015825 sec
3 imfilter 0.018062 sec
4 imfilter 0.008963 sec
5 imfilter 0.009365 sec
6 imfilter 0.010428 sec
7 imfilter 0.009566 sec
[...]

PR #607, opened 2022-01-12
Automatic event loop start/stop when not showing anything

IanButterworth:
As reported in https://github.com/JuliaGraphics/Gtk.jl/issues/503 & https://github.com/JuliaLang/julia/issues/35552 loading Gtk in a julia session basically ruins the performance of the session from there-on out, especially threading.

Fixing the core issue appears to be intrinsic to how julia runs threads, and might take a while to fix(?)

So... this PR:

- Introduces `Gtk.enable_eventloop(b::Bool)` for start/stopping the GLib event loop, which prevents blocking other julia tasks/threads
- By default starts Gtk in idle state with no event loop
- The event loop is then started with any call to `show` or `showall` so that rendering always starts when something is shown
- Then stops the eventloop when all `show`-ed widgets have been destroyed
- Previous behavior can be regained via setting env var "GTK_AUTO_IDLE" to "false", or `Gtk.AUTO_IDLE[] = false`

~~Big caveat:~~
~~I followed [this](https://www.manpagez.com/html/glib/glib-2.56.0/glib-The-Main-Event-Loop.php#GSource) for guidance on what `uv_prepare` and `uv_check` should return to set idle state, but it's all effectively just setting the timeout to 0, resulting in a cpu thread at 100%. It would clearly be better to start/stop the main thread, so it would be good to implement what @StefanMathis demoed here https://github.com/JuliaGraphics/Gtk.jl/issues/503#issuecomment-980631162~~
Update: This now fully starts/stops the eventloop, so no cpu penalty.

#### Currently execution is slowed from the moment Gtk is launched
```julia
julia> @time Threads.@threads for _ in 1:Threads.nthreads()
       sleep(0.1)
       end
  0.124272 seconds (43.64 k allocations: 2.401 MiB, 19.35% compilation time)

julia> @time using ProfileView
Gtk-Message: 00:04:43.548: Failed to load module "canberra-gtk-module"
Gtk-Message: 00:04:43.549: Failed to load module "canberra-gtk-module"
  2.335983 seconds (6.44 M allocations: 363.227 MiB, 4.19% gc time, 54.24% compilation time)

julia> @time Threads.@threads for _ in 1:Threads.nthreads()
       sleep(0.1)
       end
  5.116726 seconds (15.36 k allocations: 839.915 KiB, 0.20% compilation time)

julia> @time @profview sleep(1);
  2.141247 seconds (3.60 M allocations: 195.752 MiB, 2.05% gc time, 51.75% compilation time)

julia> @time Threads.@threads for _ in 1:Threads.nthreads()
       sleep(0.1)
       end
  5.115667 seconds (15.36 k allocations: 839.884 KiB, 0.18% compilation time)

## Gtk window closed here ##

julia> @time Threads.@threads for _ in 1:Threads.nthreads()
       sleep(0.1)
       end
  5.114526 seconds (15.36 k allocations: 839.931 KiB, 0.15% compilation time)
```

#### With this PR, execution is only slowed when the Gtk window is open and static
```julia
julia> @time Threads.@threads for _ in 1:Threads.nthreads()
       sleep(0.1)
       end
  0.123857 seconds (43.64 k allocations: 2.401 MiB, 18.54% compilation time)

julia> @time using ProfileView
Gtk-Message: 00:01:48.147: Failed to load module "canberra-gtk-module"
Gtk-Message: 00:01:48.147: Failed to load module "canberra-gtk-module"
  2.347176 seconds (6.44 M allocations: 362.962 MiB, 6.95% gc time, 52.47% compilation time)

julia> @time Threads.@threads for _ in 1:Threads.nthreads()
       sleep(0.1)
       end
  0.113058 seconds (15.35 k allocations: 839.790 KiB, 10.38% compilation time)

julia> @time @profview sleep(1);
  2.278810 seconds (3.60 M allocations: 196.749 MiB, 1.81% gc time, 49.04% compilation time)

julia> @time Threads.@threads for _ in 1:Threads.nthreads()
       sleep(0.1)
       end
  5.116526 seconds (15.36 k allocations: 839.884 KiB, 0.18% compilation time)

## Gtk window closed here. ##

julia> @time Threads.@threads for _ in 1:Threads.nthreads()
       sleep(0.1)
       end
  0.110775 seconds (15.35 k allocations: 839.228 KiB, 8.27% compilation time)

julia> @time Threads.@threads for _ in 1:Threads.nthreads()
       sleep(0.1)
       end
  0.110090 seconds (15.35 k allocations: 839.790 KiB, 8.71% compilation time)

julia> @time Threads.@threads for _ in 1:Threads.nthreads()
       sleep(0.1)
       end
  0.109473 seconds (15.35 k allocations: 839.790 KiB, 8.32% compilation time)
```

## Notes

1) Note that the blocking still happens while windows are open and static, so this is just a bandaid.

### DIFF
--- a/src/GLib/GLib.jl
+++ b/src/GLib/GLib.jl
@@ -32,6 +32,7 @@ export set_gtk_property!, get_gtk_property
 export GConnectFlags
 export @sigatom, cfunction_
 
+const IDLE = Ref{Bool}(true)
 
 cfunction_(@nospecialize(f), r, a::Tuple) = cfunction_(f, r, Tuple{a...})
 
--- a/src/GLib/GLib.jl
+++ b/src/GLib/GLib.jl
@@ -32,7 +32,6 @@ export set_gtk_property!, get_gtk_property
 export GConnectFlags
 export @sigatom, cfunction_
 
-const IDLE = Ref{Bool}(true)
 
 cfunction_(@nospecialize(f), r, a::Tuple) = cfunction_(f, r, Tuple{a...})
 
@@ -41,6 +40,8 @@ cfunction_(@nospecialize(f), r, a::Tuple) = cfunction_(f, r, Tuple{a...})
         @cfunction($(Expr(:$,:f)), $rt, ($(at.parameters...),))
     end
 end
+
+const gtk_eventloop_f = Ref{Function}()
 
 # local function, handles Symbol and makes UTF8-strings easier
 const  AbstractStringLike = Union{AbstractString, Symbol}
--- a/src/GLib/signals.jl
+++ b/src/GLib/signals.jl
@@ -288,7 +288,9 @@ function uv_prepare(src::Ptr{Nothing}, timeout::Ptr{Cint})
     global expiration, uv_pollfd
     local tmout_ms::Cint
     evt = Base.eventloop()
-    if !_isempty_workqueue()
+    if IDLE[]
+        return Int32(1)
+    elseif !_isempty_workqueue()
         tmout_ms = 0
     elseif !uv_loop_alive(evt)
         tmout_ms = -1
@@ -316,7 +318,7 @@ end
 function uv_check(src::Ptr{Nothing})
     global expiration
     ex = expiration::UInt64
-    if !_isempty_workqueue()
+    if !_isempty_workqueue() || IDLE[]
         return Int32(1)
     elseif !uv_loop_alive(Base.eventloop())
         return Int32(0)
--- a/src/GLib/signals.jl
+++ b/src/GLib/signals.jl
@@ -288,9 +288,7 @@ function uv_prepare(src::Ptr{Nothing}, timeout::Ptr{Cint})
     global expiration, uv_pollfd
     local tmout_ms::Cint
     evt = Base.eventloop()
-    if IDLE[]
-        return Int32(1)
-    elseif !_isempty_workqueue()
+    if !_isempty_workqueue()
         tmout_ms = 0
     elseif !uv_loop_alive(evt)
         tmout_ms = -1
@@ -318,7 +316,7 @@ end
 function uv_check(src::Ptr{Nothing})
     global expiration
     ex = expiration::UInt64
-    if !_isempty_workqueue() || IDLE[]
+    if !_isempty_workqueue()
         return Int32(1)
     elseif !uv_loop_alive(Base.eventloop())
         return Int32(0)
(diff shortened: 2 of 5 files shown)


tknopp:
Thanks Ian for pushing this forward, this is much appreciated!!!

I would like to ping @jpsamaroo @tkf @vtjnash @JeffBezanson @chriselrod: Can anybody from you give a perspective on this PR or what the route forward for proper Gtk mainloop integration with Julia Threads/Tasks is? It is really a problem that one cannot profile multi-threading applications with `ProfileView`.

tknopp:
@IanButterworth: Ping me once you think this is ready and can be merged + release. A workaround a certainly better than nothing right now.

IanButterworth:
I'm happy with this now. Tests pass locally. Windows open and behave correctly, including with ProfileView, with no downstream, code changes.

@tknopp one question is whether there are any more calls to display widgets/windows that need the `handle_auto_idle(w)` function added, to ensure the eventloop is started?

jpsamaroo:
@tknopp PR https://github.com/JuliaLang/julia/pull/42302 implements threadpools, which would allow Gtk.jl to request that some subset of its tasks get scheduled in an isolated threadpool. Those tasks would be isolated from the rest of the Julia runtime (and thus achieve lower latency even under higher compute load from unrelated code). If no additional threadpools are configured, the tasks will instead remain in the primary threadpool.

I'm implementing the Julia API for threadpools now, and will let you know once the PR is ready for testing!

IanButterworth:
Perhaps even when the threadpools work is released in base, this PR might remain a good idea? 

Someone who just wants to profile their multithreaded system with Gtk probably won't want Gtk to swallow a thread while the profile is being collected?

StefanMathis:
I do think this PR is a good idea even with thread pools (which I also think are a great idea). When I don't have an open Gtk window, I definitely want all cores to contribute as much as possible to e.g. a multithreaded loop. This means that the Gtk main loop shouldn't be active at all unless it is actually needed, which is guaranteed by this PR.

IanButterworth:
I did find what appears to be a redundant show in ImageView.jl that made this PR think something was open after all windows had been closed though. Not quite sure what's happening there https://github.com/JuliaImages/ImageView.jl/pull/259

But I don't think that needs to hold this back. There may be tweak PRs following this to catch all show entry and exit points.

tknopp:
So I am for merging this even if thread pools come in the future, which will not before 1.8 anyway. @IanButterworth: Is this good to go from your side?

IanButterworth:
Indeed. Good to go as a non-breaking change, and I'll watch out for downstream bugs.

timholy:
This is a great idea, thanks so much for fixing this vexing problem!

PR #611, opened 2022-01-16
Add a yield for windows (fix #610). Fix window not finishing close on MacOS

### LINKED ISSUE #610: Dialog windows sometimes not doing anything on v1.1.11
The changes introduced in PR #607 seem to cause issues when showing dialog boxes sometimes, at least on Windows. Setting `GTK_AUTO_IDLE` to `false` fixes it, confirming that this PR is the culprit.

We don't have much information to present unfortunately, but in some situations, opening a dialog box such as `ask_dialog` shortly after `using Gtk` opens a completely unresponsive window. It's not possible to interact with any buttons in the window, including the X button in the window bar. In our tests, adding a simple `sleep(0)` or `println()` in front of the `ask_dialog` is already enough to fix it, so we don't really know why this could happen.

IanButterworth:
Fixes #610
The fix indicated to work for @Vexatos https://github.com/JuliaGraphics/Gtk.jl/issues/610#issuecomment-1013760658
I'm not aware of this needing to be done outside of Windows, so decided to be conservative.

Also fixes an issue on MacOS where the window doesn't always close and freezes. 
It seems like there's a sequencing bug on MacOS, as it's pretty normal to call `quit_main` on `:destroy`
The async wait just gives some time for the window to finish closing.

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -39,6 +39,7 @@ function handle_auto_idle(w::GtkWidget)
                 isempty(shown_widgets) && enable_eventloop(false)
             end
         end
+        Sys.iswindows() && yield() # issue #610
     end
 end
 function show(w::GtkWidget)
--- a/src/base.jl
+++ b/src/base.jl
@@ -34,7 +34,7 @@ function handle_auto_idle(w::GtkWidget)
         signal_connect(w, :realize) do w
             enable_eventloop(true)
             shown_widgets[w] = nothing
-            signal_connect(w, :destroy) do w
+            signal_connect(w, :destroy, #= after =# true) do w
                 delete!(shown_widgets, w)
                 isempty(shown_widgets) && enable_eventloop(false)
             end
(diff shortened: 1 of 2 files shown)


src/base.jl
```diff
@@ -39,6 +39,7 @@ function handle_auto_idle(w::GtkWidget)
                 isempty(shown_widgets) && enable_eventloop(false)
             end
         end
+        Sys.iswindows() && yield() # issue #610
```
giordano, line 42: Make it `@static`?

tknopp:
Would it hurt to make the yield also on other platforms? It might be that this issue also happens on other platforms. Maybe just more rarely since it is some timing/scheduling issue.

Anyway I merge this now since it fixes the issue. Thank you very much Ian.

IanButterworth:
> Would it hurt to make the yield also on other platforms?

Definitely worth trying if there are reports of the same thing on other platforms

Issue #609, opened 2022-01-14
`showall` overrides controls marked with `visible == false`


BioTurboNick:
`showall` is the method used to render the controls added to a Gtk window.

However, it also overrides any controls that had been set to `visible == false`. This is an issue when trying to design a GUI with some default-invisible controls in Glade, as they have to be turned off again after `showall`.

Now, this behavior is possibly obvious from the name `showall`, but perhaps a different method is needed for the initial display call that doesn't have that behavior?

tknopp:
- is this used in Gtk.jl itself or just in the tests?
- could you try just calling `show` on a window and look what happens?
- look what typical Gtk examples do?

BioTurboNick:
It's used all over the documentation. E.g. https://juliagraphics.github.io/Gtk.jl/latest/manual/gettingStarted/

When I just call `show` on the window, some elements aren't rendered.

jwahlstrand:
There is a "no_show_all" property that lets a widget ignore "showall" calls to its parents. It also can be set within Glade.

tknopp:
That it is used in the documentation does not mean a lot because `showall` is a valid Gtk function. My question was rather, if we abuse it somewhere in Gtk.jl internally where it should not be used and surprisingly makes widgets visible.

BioTurboNick:
Yeah, I guess it's just a part of Gtk's design I didn't appreciate. I'll find a spot in the documentation to mention this.

BioTurboNick:
But to answer your question, no it's not an issue inside Gtk.jl. It's only used once and in that location it's appropriate.

PR #612, opened 2022-01-16
Add documentation to showall

BioTurboNick:
Specifically, calling out how it interacts with `visible` property and how `no_show_all` can be used to prevent that.

Addresses #609


tknopp:
perfect.

PR #615, opened 2022-01-17
Add `wait_eventloop_stopping()` and wait before recording state before a pause

IanButterworth:
If the eventloop was still async stopping, `pause_eventloop()` was recording the state as running, then restarting it after execution, so wait.

Also add `Gtk.wait_eventloop_stopping()` to wait for the stop to complete.

See https://github.com/JuliaGraphics/Gtk.jl/issues/503#issuecomment-1014896142 for discussion


IanButterworth:
Tests are passing but I'm still exploring this PR.

It would be nice to remove the async eventloop stop, but I'm still getting windows staying open and freezing on macos.

PR #621, opened 2022-02-02
Add Graphite2_jll as a dependency to fix CI fail

jdeldre:
This PR is an attempt to address the failure of the CI on the Julia nightly build, which seems to center on the `Graphite2_jll` artifact. The error comes up on the latest CI for the package, including [here](https://github.com/JuliaGraphics/Gtk.jl/runs/4914062539?check_suite_focus=true). In this PR, `Graphite2_jll` is added as an explicit dependency.


giordano:
I don't understand how this solves anything. I'm not even sure the reported error is real, as opposed to a random network failure

jdeldre:
> I don't understand how this solves anything. I'm not even sure the reported error is real, as opposed to a random network failure

It has been repeatedly breaking CI for a week or so on the nightly build for my [own](https://github.com/JuliaIBPM/ImmersedLayers.jl) package because of its dependence on `Gtk.jl`.

giordano:
But graphite2_jll is already a transitive dependency, how does adding it to the project change anything?  If there is an issue, it's somewhere else.

jdeldre:
Agreed, this was a desperation kluge. On deeper inspection it seems to be an issue with the most recent version of graphite2_jll (1.3.14) on the nightly build of Julia. I've managed to isolate it and I've opened an issue here JuliaPackaging/Yggdrasil#4358

PR #622, opened 2022-02-06
Support `deleteat!` for all `GtkContainer`s

GoldenSnitchD:
Found `GtkButton`s were not supported to `deleteat!` before.
Now add it.


tknopp:
LGTM

PR #625, opened 2022-02-15
Change property handling, generate objects, interfaces, some structs using GI, and generate getters and setters using GI

jwahlstrand:
This modifies how properties are handled, consistent with a discussion with @tknopp (#602). Properties can be accessed and set using, for example, `w.props.title`. This is the same as in Python. However, autocomplete in the REPL doesn't work (https://discourse.julialang.org/t/foo-bar-tab-completion-not-working/61052). So now I'm leaning toward instead mapping GObject properties directly onto julia properties, so that we can use `w.title`. The only potential downside I can think of: subclassed GObjects would have to avoid using field names that are the same as properties.

It also includes getter and setter methods generated by GI.jl, which take the form, e.g., `get_title(w::GtkWindow)` and `set_title(w::GtkWindow, title::AbstractString)`. These are exactly what is in GAccessor but should behave better in some ways. For example, they accept `nothing` when the C function accepts `C_NULL`, they return `nothing` rather than `C_NULL` (or throwing an "undefined reference" error), and they return strings rather than `Ptr{UInt8}`. This PR currently does nothing to GAccessor but it should be straightforward to replace it with something that calls the methods defined here. I would like to eventually use GI to export all methods and functions from libgtk and GObject-based dependencies. Maybe these auto-generated methods should live in their own submodules so that users know they're different. For example, some methods might need to be called using `@sigatom` and I don't see a way of automatically marking them.

Finally, it uses GI.jl to define GObject, GInterface, and (opaque) GBoxed types. It defines all of these in libgtk and a selection from Pango, Gdk, and GLib. I modified some of the code in gtype.jl to support the GI "transfer" annotations, which prevent memory leaks in GI-generated methods. The default is "transfer=none", which should behave exactly like the old code (that was my intention, at least). I have verified that in a few cases this does plug memory leaks, but problems will probably arise and transfer handling will probably have to be refined. Handling of GBoxed types in GI.jl is a bit different than what's currently in Gtk.jl and needs more testing. 

Non-opaque GBoxed types, where there are accessible fields, are not dealt with here. These include important stuff like `GtkTreeIter` and `GtkTextIter`. The issue is basically the following. For non-opaque structs, we export structs as `GtkStruct`, which has a handle that points to the C struct. But for structs with fields, we'd like to map the C struct onto a Julia struct. In some places in Gtk.jl, this is done by defining an underscore version, `_GtkStruct` (see for example `_GList`). But in most other cases this isn't done (`GValue`, `GtkTreeIter`, `GtkTextIter`). So to make this consistent we would have to do some renaming. I would lean toward being consistent and using the underscored name for the struct with fields.

@tknopp, you mentioned creating a "v2" branch to collect breaking changes like this. This could be a starting point for that. Or should v2 also be a switch to GTK4? Maybe it would be easier on dependencies to leave this package alone and make a new one for GTK4...


tknopp:
First comment: I have had great experience in NFFT.jl in grouping together different packages:

   https://github.com/JuliaMath/NFFT.jl

You can see that I have four packages in just one Github repo and this makes maintaining things IMHO much easier:

NFFTPackages

So I would propose that we start with one repo where you do the restructuring.

(Here is the Latex tree generator: https://lautgesetz.com/latreex/)

Regarding Gtk3/4: How do other handle this? I see the following options
1. Try to support both Gtk3/4 in a single package. Just works if the API is 99% the same, otherwise it gets ugly.
2. Drop Gtk3 and let Gtk always follow the most recent Gtk version
3. Make two packages Gtk3 and Gtk4 within the same Github repo. They can use same GtkBase package, i.e. share infrastructure.

Ping: @timholy @jonathanBieler @giordano:    Jared needs our guidance and this is an important decision where I don't want to drive things into the wrong direction. @giordano: Your input from BinaryBuilder land would also be very imortant since we lay out the foundation for Gtk3,4,5,... 

Thanks!

tknopp:
And whether to use a branch or start a new Github repository depends a little bit how things are structured at the end. If we have all packages in a single Github (nice for testing) then we can think of abandoning https://github.com/JuliaGtk and just have a single Gtk repo in JuliaGraphics. Otherwise it would make sense grouping `G*` things in https://github.com/JuliaGtk. I don't have strong opinions here.

jwahlstrand:
It seems like the jump from Gtk 3 to 4 is bigger than Gtk 2 to 3, and it would be pretty messy to support both in the same module. For example, there is no GtkContainer in Gtk4, events are handled using "event controller" objects, and drawing in custom widgets was changed a lot too.

I think it makes the most sense to simultaneously switch to Gtk4 and use GObject introspection. It would cause needless suffering to break everyone's code with GI, then break it again with Gtk4. So maybe I'll focus on Gtk4 going forward, using this general approach (comments and suggestions would be welcomed). Once that works we will see how painful it will be to port code to use it (and how well it works across platforms) and we can decide which of your options to pick.

I like your idea of keeping all dependencies in one repo.

tknopp:
Then probably the best is to start https://github.com/JuliaGtk/Gtk4 and start fresh. If we decide to just call it just`Gtk` in the end we can simply do so later. Calling it `Gtk4` for the moment has the advantage that one can install/develop both packages in parallel.

You should have the power to create repositories in https://github.com/JuliaGtk/

jwahlstrand:
Sounds good!

giordano:
I just wanted to mention that I created `GTK4_jll` as a separate package from `GTK3_jll` so that they can be used together in the same environment, at least in principle.  Allowing libgtk4 to sit next to libgtk3 is also what Linux distributions do.

tknopp:
Thanks Mose, this underlines that making a `Gtk4` package is a good path forward. This gives us maximum flexibility in the transition period.

tknopp:
@jwahlstrand: Thanks for moving `Gtk4`! What about [GI.jl](https://github.com/jwahlstrand/GI.jl)? Do you want to make that a sub package  within the `Gtk4` repository or keep it separate? In the later case you can move it to JuliaGtk as well so that we have all within one place.

jwahlstrand:
I'm thinking it should be a subpackage. It uses some GLib functions but also generates functions for GLib. It's going to take a bit of time to get everything consolidated and working again.

tknopp:
Cool, this is exactly that sort of situation where a sub package in my opinion shines. If you have questions, please ping me. NFFT.jl is a good template IMHO. We can proceed the discussions in the Gkt4 repository.

PR #628, opened 2022-02-21
add option for "simple" main loop iteration

jwahlstrand:
I hesitated to post this because it seems too good to be true. This PR creates a "simple" main loop option (disabled by default) that replaces the call to `gtk_main()` with a Timer that iterates the GLib main loop every 5 ms in a non-blocking way. It also skips the call to `__init__gmainloop__()`, completely omitting the creation of a GSource for libuv. This is enabled using GTK_SIMPLE_LOOP = true.

When I run Julia 1.7 with `GTK_SIMPLE_LOOP=true GTK_AUTO_IDLE=false JULIA_NUM_THREADS=4`, I don't see a slowdown in the multithread tests posted at the beginning of #503. All Gtk.jl tests pass for me locally on Linux and Windows. ProfileView, ImageView and Winston seem to work just fine.

Am I missing something? It's very possible I just haven't managed to create the situations that led to the tight integration of the libuv loop with the GLib loop. On the other hand, that code is ~8 years old and maybe isn't needed in recent versions of Julia?

### DIFF
--- a/src/GLib/signals.jl
+++ b/src/GLib/signals.jl
@@ -404,7 +404,7 @@ const simple_loop = Ref{Bool}(false)
 
 const exiting = Ref(false)
 function __init__()
-    simple_loop[] = get(ENV, "GTK_SIMPLE_LOOP", "false") == "true"
+    simple_loop[] = get(ENV, "GTK_SIMPLE_LOOP", "true") == "true"
     if isdefined(GLib, :__init__bindeps__)
         GLib.__init__bindeps__()
     end
--- a/src/GLib/signals.jl
+++ b/src/GLib/signals.jl
@@ -400,8 +400,11 @@ macro idle_add(ex)
     end
 end
 
+const simple_loop = Ref{Bool}(false)
+
 const exiting = Ref(false)
 function __init__()
+    simple_loop[] = get(ENV, "GTK_SIMPLE_LOOP", "false") == "true"
     if isdefined(GLib, :__init__bindeps__)
         GLib.__init__bindeps__()
     end
@@ -410,6 +413,8 @@ function __init__()
     exiting[] = false
     atexit(() -> (exiting[] = true))
     __init__gtype__()
-    __init__gmainloop__()
+    if !simple_loop[]
+        __init__gmainloop__()
+    end
     nothing
 end
--- a/src/Gtk.jl
+++ b/src/Gtk.jl
@@ -174,6 +174,15 @@ function __init__()
     enable_eventloop(!auto_idle[])
 end
 
+function iteration(timer)
+    ccall((:g_main_context_iteration, libglib), Cint, (Ptr{Cvoid}, Cint), C_NULL, false)
+end
+
+function glib_main_simple()
+    t=Timer(iteration,0.01;interval=0.005)
+    wait(t)
+end
+
 const auto_idle = Ref{Bool}(true) # control default via ENV["GTK_AUTO_IDLE"]
 const gtk_main_running = Ref{Bool}(false)
 const quit_task = Ref{Task}()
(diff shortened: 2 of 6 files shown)


tknopp:
So instead of running the gtk main loop on its own long-running task this just runs one iteration of the main loop in a dedicated task. I did not know that GLib offers that functionality and it seems to make more sense for Gtk.jl since it offers more fine-grained control.

jwahlstrand:
Yes, this would also let us suspend the loop momentarily without quitting it. But critically this doesn't seem to disturb multithreading at all. My concern is that this skips >100 lines of code in signals.jl. Is that really not needed? It would be good for others to test this.

jwahlstrand:
Two of the three failures in the previous CI run were caused by a canvas widget not being realized in time on macos for a call to getgc(). Modiying a test to force an iteration of the loop solves that. The third was a segfault and that makes me nervous... I will continue to putter around with this idea.

tknopp:
I am not 100% sure but probably the while loop should move into `getgc`?

jwahlstrand:
Yeah, maybe one could iterate the loop if the widget isn't realized in `getgc`.

Skipping the call to `__init__gmainloop__` disables all the `g_yield` machinery, which is used by `@sigatom`. That seems important. However, in the docs at the bottom of https://juliagraphics.github.io/Gtk.jl/latest/manual/signals/, there's a discussion of why and when you need `@sigatom`, followed by, "These restrictions should be fixed once https://github.com/JuliaLang/julia/pull/13099 is merged." That happened a long time ago, which fits the hypothesis that it's no longer necessary, but I'm definitely in over my head here.

When I have time I'll refine this PR. We'll see if that random segfault reappears.

tknopp:
Did you still see segfaults?

tknopp:
And yes after https://github.com/JuliaLang/julia/pull/13099 was merged I was able to remove all the `@sigatom`s from my (larger) Gtk application. IMHO all this can be removed.

jwahlstrand:
No, that's 4 or 5 in a row with no segfaults. But I just noticed that I broke autoidle. Once that is fixed I think this is ready to go -- I don't have time to work on it now. Do you think this should be an option for now or the new default main loop behavior?

tknopp:
If it works better, I would make it the default.

tknopp:
Whether to remove the old code, I am not entire sure. Probably it would be better to make the cleanup in the Gtk4.jl project?

jwahlstrand:
I tried this out on Windows this evening and, despite all tests passing, it's a disaster. Resizing windows and interacting with the GUI is very sluggish. It works fine on Linux...

tknopp:
The windows issue is of course very unfortunate. Is this completely solved by disabling the client side decorations?

Probably the best would be to keep both possibilities (as you do now) and still cleanup the old path (`@sigatom`...).

tknopp:
Any performance degradation worries me. The experience on OSX is not so great. But as long as we have the switch, its fine I think.

IMHO we should look forward to the upcoming Gtk4 experience and put most energy there.

> OK, do you think we should clean up the @sigatom stuff here? I could at least remove the mention in the docs. I need to add some words to the docs about how to switch between this and auto idle, but after that I think it's ready to go.

I don't care if you still remove it here or at the Gtk4 repository. My comment was more targeting the later: I.e. that the Gtk4 code base can clean this up. So probably the best is to leave it here as is, so that we can merge the PR and then you can carry over a cleaned up version to Gtk4.

jwahlstrand:
With this last commit, the performance issues on Mac and Windows are fixed. Disabling client side decorations was still necessary on Windows because without that there is a constant spew of `Gdk-WARNING **: gdk-frame-clock: layout continuously requested, giving up after 4 tries`. I think that we should go with #630, however, because it's a more elegant solution.

tknopp:
what should we do with this? Keep the infrastructure for the future to remain flexible?

jwahlstrand:
Nah, this is sort of a mess.

PR #630, opened 2022-03-02
Disable uv_backend_fd integration

vtjnash:
Might fix https://github.com/JuliaGraphics/Gtk.jl/issues/503
Closes https://github.com/JuliaLang/julia/issues/35552


jwahlstrand:
I can confirm that this works on Mac, Linux, and Windows. I see some sporadic lag in the REPL but it's not really reproducible or particularly bothersome. Thanks for fixing this.

PR #635, opened 2022-03-25
Document "`key-release-event"`

JeffFessler:
This PR adds an example of capturing key release events, to augment the previous key press example.

This particular example is motivated by the use case of using a computer keyboard
like a piano keyboard where you want to know the duration the key was pressed,
but surely there are other uses given that GTK does support key release events.


JeffFessler:
@tknopp you were the author of the original doc here.  Small world :)

tknopp:
looks good to me, thanks!

PR #636, opened 2022-03-27
Fix async UI example

JeffFessler:
The Asynchronous UI example does not run as presented in the docs
because `addprocs` and `@fetchfrom` are now in `Distributed`,
so this PR addresses that issue.

Along the way, this PR also fully qualifies all the methods that are used from `Gtk`
so that is clear to new users (and old users like myself with limited memory)
where functions originate, especially useful for those with general names like `start` and `showall`.

The text at the top of these docs says that "Julia has currently no possibility of multithreading"
which seems likely to be an outdated claim, in light of `Threads.@threads` etc.
Before merging this PR, please reword those lines (maintainers can edit) or offer advice on how to reword.  I wasn't quite sure how to reword because I don't fully understand this example.

### DIFF
--- a/docs/src/manual/async.md
+++ b/docs/src/manual/async.md
@@ -1,11 +1,14 @@
 # Asynchronous UI
 
-Here is an example of an asynchronous update of the user interface. Since
-Julia has currently no possibility of multithreading we use a second process
-to offload the work. The example is just a proof of principle.
+Here is an example of an asynchronous update of the user interface.
+Since Julia has currently no possibility of multithreading
+we use a second process to offload the work.
+The example is just a proof of principle.
 
 ```julia
-using Gtk
+using Gtk: GtkButton, GtkSpinner, GtkEntry, GtkGrid, GtkWindow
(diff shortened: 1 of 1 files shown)


tknopp:
That example is pretty obsolete now. These discourse threads contain information how one should nowadays do:

https://discourse.julialang.org/t/threading-1-3-success-story/27111/2
https://discourse.julialang.org/t/gui-gtk-jl-ui-freezing-when-signal-connect-used/75156
https://discourse.julialang.org/t/ensuring-gtk-jl-responsivity/40710

My advice is:
- the callback should directly spawn a worker thread.
- the worker thread uses `idle_add` when needing to change UI elements.

To prevent the background thread to run on the first thread and block UI one can use the package https://github.com/tro3/ThreadPools.jl where I usually use `@tspawnat` to spawn the task on the second thread. Unfortunately one still needs to startup Julia with `julia -t 2`. No way to increase the number of threads at runtime.

JeffFessler:
Thanks for the helpful feedback!  That's more involved than I can handle right now, so I'm just going to report it as an issue #640 and hopefully someone will eventually update this PR accordingly.

PR #638, opened 2022-03-30
Add slider (`GtkScale`) example

JeffFessler:
This example is in response to students asking "how to" questions about sliders.
Hopefully it will help other beginners too.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -21,6 +21,7 @@ makedocs(
                      "manual/keyevents.md",
                      "manual/canvas.md",
                      "manual/customWidgets.md",
+                     "manual/slider.md",
                      "manual/async.md",
                      "manual/nonreplusage.md",
                      "manual/packages.md"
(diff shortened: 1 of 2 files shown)


timholy:
Beginners should perhaps consider GtkObservables since you don't really need to know much about Gtk internals. A simplified version of your example is
```julia
using GtkObservables
slider1 = slider(0:10)
slider2 = slider(0:30; value = slider1[])
on(slider1) do val
    slider2[] = val
end
```
(obviously these have to be placed in a window).

tknopp:
> Beginners should perhaps consider GtkObservables since you don't really need to know much about Gtk internals. 

Probably. But let's please not mix Gtk.jl and GtkObservable.jl documentation. My vision would be to keep Gtk.jl low-level and close to the C-lib and have high-level code in an additional package such as GtkObservables.

timholy:
Agreed 100%. Just pointing it out here.

JeffFessler:
Ok, to avoid mixing documentation yet still nudge beginners (like me!) the easier way, I made a separate PR #639 to add a link to it in the documentation.  (Hopefully a single link is too much mixing!)
I made it a separate PR because they are independent and maybe you will accept one but not the other.
Anything else I need to do on this one?

JeffFessler:
bump :)

PR #656, opened 2022-08-03
Update async.md for 1.0 and Threading

### LINKED ISSUE #640: Async UI example needs updating
The example in this file needs updated:
https://github.com/JuliaGraphics/Gtk.jl/blob/master/docs/src/manual/async.md

I made a version that at least runs here: #636, but apparently there are even better ways to do it.

LilithHafner:
Closes #636
Closes #640

### DIFF
--- a/docs/src/manual/async.md
+++ b/docs/src/manual/async.md
@@ -1,11 +1,13 @@
 # Asynchronous UI
 
-Here is an example of an asynchronous update of the user interface. Since
-Julia has currently no possibility of multithreading we use a second process
-to offload the work. The example is just a proof of principle.
+It is possible to perform background computation without interfering with user interface
+responsiveness either using separate processes or using multithreading. Use of a separate
+process includes slightly more overhead but is also more robust.
+
+Here is an example using a separate process to offload the work.
 
 ```julia
-using Gtk
+using Gtk, Distributed
 
 btn = GtkButton("Start")
 sp = GtkSpinner()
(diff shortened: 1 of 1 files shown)


tknopp:
I would actually remove the multi-process version.

LilithHafner:
Really? When running the default `julia -t1` the Distributed approach works for me and the Threads approach stalls the UI.

tknopp:
Because ist just works in toy examples for me. If the callback needs to call package code one needs to run Julia with -p and "@everywhere using ...". Data exchange ist also more complicated. For me the multi-threaded version is much simpler.

If I see it right, Julia 1.8 will give us the second thread by default.

LilithHafner:
> If the callback needs to call package code one needs to run Julia with -p and "@Everywhere using ...". Data exchange ist also more complicated.

Noted. Threading certainly has advantages and should probably be listed first.

> If I see it right, Julia 1.8 will give us the second thread by default.

Where did you see that?

```julia
x@X ~ % julia +beta
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _' |  |
  | | |_| | | | (_| |  |  Version 1.8.0-rc3 (2022-07-13)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> Threads.nthreads()
1
```

In 1.9 we will have interactive threads which should make the threading approach more robust:

"Threads.@spawn now accepts an optional first argument: :default or :interactive. An interactive task desires low latency and implicitly agrees to be short duration or to yield frequently. Interactive tasks will run on interactive threads, if any are specified when Julia is started ([#42302])." (https://github.com/JuliaLang/julia/blob/df846436bb91b291be988f5e18e256d514b97347/NEWS.md)

I think the best option is probably to recommend threads and then afterward, note that Distributed is also an option with caveats.

tknopp:
🙁 I thought this would be in 1.8.

But anyway. I like your suggestion of just switching the order.

tknopp:
thanks!

PR #681, opened 2022-12-26
make sure value wrapped in `mutable` is not garbage collected

KristofferC:
See discussion in https://github.com/JuliaLang/julia/pull/47184#issuecomment-1365404104

cc @vtjnash


vchuravy:
I think the correct fix is to define these in `unsafe_convert` instead of `cconvert` https://github.com/JuliaGraphics/Gtk.jl/blob/57c3bfe852afcedd0e781ffc7703e811114b30b2/src/GLib/MutableTypes.jl#L37-L38

KristofferC:
Feel free to update, otherwise I'll do it in the morning.

timholy:
Incorporated into #682

PR #692, opened 2023-08-10
turn off auto-idle by default on Windows

jwahlstrand:
I think this is the right default for Windows. The lag caused by `using Gtk` on Windows is very frustrating. See https://github.com/timholy/ProfileView.jl/issues/205


timholy (approved):

Vexatos:
Not sure whether this will fix all of the keyboard lag, as the changed line was introduced in #607 in 2022 while the keyboard input issue on Windows has been known since at least 2018.

PR #697, opened 2024-01-12
Allow Gtk.jl to be installed on Windows on Julia 1.10.

nhz2:
This PR allows `Gtk.jl` to be installed on Windows on Julia 1.10. 

The version of `adwaita_icon_theme_jll` required by `Gtk.jl` contains symlinks. This prevents it from being installed on Windows in Julia 1.10 due to stricter artifact integrity checking. Ref: https://github.com/JuliaLang/Pkg.jl/issues/3643 and https://github.com/JuliaPackaging/Yggdrasil/issues/7860

This got fixed for the most recent version of `adwaita_icon_theme_jll` in https://github.com/JuliaPackaging/Yggdrasil/pull/7930

However, it seems like `adwaita_icon_theme_jll` version 43 is not compatible with `Gtk.jl` Ref: https://github.com/JuliaGraphics/Gtk.jl/pull/696

This PR replaces the `adwaita_icon_theme_jll` dependency with a regular artifact. For Linux and BSD this doesn't change anything, for Windows, a special version of `adwaita_icon_theme` is installed where the symlinks have been replaced with copies. This Windows version has the same tree hash as the version installed in Julia 1.9 on Windows if symlinks cannot be made. 

You can check the `git-tree-sha1` values in the `Artifact.toml` in this PR
and compare to the reported expected and calculated hashes in https://github.com/JuliaPackaging/Yggdrasil/issues/7860
```
Calculated git-tree-sha1: 043d0732f572e66059278f5964e3752f018a5cdf
git-tree-sha1 = "043d0732f572e66059278f5964e3752f018a5cdf"
```

Also, currently, I am storing the Windows artifact at `https://github.com/medyan-dev/SmallZarrGroups.jl/releases/download/v0.6.6/copy_symlinks_adwaita_icon_theme.v3.33.92.any.tar.gz` but there may be a better place for it.

CC @jwahlstrand @revolutionflourlake772


nhz2:
Closing because I think https://github.com/JuliaPackaging/Yggdrasil/pull/8003 is a better fix.

